PR #2, opened 2023-01-10
NR-75385 feat: first NR distro structure

rubenruizdegauna:


configs/README.md
```diff
@@ -0,0 +1,3 @@
+This file contains the default configuration files for the container images. Due to a problem with goreleaser, these cannot be in the a directory with the same name as the binary for that image. For intance, if the binary is `opentelemetry-collector`, the container image cannot include anything from a directory named `opentelemetry-collector`, which is exactly what we have.
```
rogercoll, line 1:
```suggestion
This directory contains the default configuration files for the container images. Due to a problem with goreleaser, these cannot be in the a directory with the same name as the binary for that image. For intance, if the binary is `opentelemetry-collector`, the container image cannot include anything from a directory named `opentelemetry-collector`, which is exactly what we have.
```?

File: configs/nr-otel-collector.yaml [hunk 2]
```diff
@@ -0,0 +1,57 @@
...
+
+extensions:
+  health_check:
+  pprof:
+    endpoint: 0.0.0.0:1777
+  zpages:
+    endpoint: 0.0.0.0:55679
+
+receivers:
+  otlp:
```
rogercoll, line 11: Should we add the default configuration of the hostmetrics receiver?

rogercoll (commented):
Just two small questions, other than that, LGTM

configs/nr-otel-collector.yaml
rubenruizdegauna: as we discussed we need to decide on the default config. Let's leave it as it is for now (with the TODO https://github.com/newrelic/opentelemetry-collector-releases/pull/2/files#diff-fc66cb839ba9e9a13bbe1d2c9ba7aa65b3b9c917c7563381511ff4f15057fb09R1)

rogercoll (approved):
LGTM

rogercoll (approved):
Nice! 🔥

PR #7, opened 2023-01-20
added docker manifest release

cristianciutea:
This PR will add the feature of creating docker images tagged as -rc on pre-release and later on the release a new manifest would be pushed without the -rc

The approach is to let goreleaser publishing the docker images and manifest with -rc suffix

Later on the release a new manifest would be pushed without -rc suffix

### DIFF
--- a/.github/workflows/ci-prerelease.yml
+++ b/.github/workflows/ci-prerelease.yml
@@ -37,11 +37,11 @@ jobs:
       - name: Generate distribution sources
         run: make generate-sources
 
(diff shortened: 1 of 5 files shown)


rogercoll:
LGTM, just a small question, should we remove rc images after release?

rogercoll (approved):
LGTM, as discussed, rc won't be removed.

rubenruizdegauna (approved):
LGTM

PR #9, opened 2023-01-25
NR-78045 Added publishing workflow

cristianciutea:
This PR will add the workflows to publish packages to s3 bucked
- on pre-release will use staging s3 bucket
- on release will use production s3 bucket


rubenruizdegauna (approved):
nice!

PR #10, opened 2023-01-30
NR-78042: add packaging tests

rogercoll:
Execution example: https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/4042601422

### DIFF
--- a/.github/workflows/ci-prerelease.yml
+++ b/.github/workflows/ci-prerelease.yml
@@ -49,7 +49,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   publishing-to-s3:
-    name: Publish artifacts into s3 staging bucket
+    name: Publish artifacts into s3 testing bucket
     uses: ./.github/workflows/component_publish.yml
     needs: [prerelease]
     secrets:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-prerelease.yml
```diff
@@ -70,3 +70,22 @@ jobs:
...
+  packaging-tests:
+    runs-on: ubuntu-20.04
+    needs: [publishing-to-s3]
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Set and validate distribution name and go-version
+        run: .github/workflows/scripts/set_version.sh
```
rubenruizdegauna, line 83: set_version is executed on pre-release step. Is it necessary here ☝️ ?

.github/workflows/ci-prerelease.yml
```diff
@@ -70,3 +70,22 @@ jobs:
...
+
+      - uses: newrelic/pkg-installation-testing-action@v1
+        with:
```
rubenruizdegauna, line 86: Is there any reason not to run the `component_molecule_packaging.yml`?

.github/workflows/ci-prerelease.yml
rogercoll: I think so, `packaging-tests` is another job (it could run on another instance) so environment variables set on the previous jobs do not persistent. I found this issue on the initial commints: https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/4024935623/workflow

.github/workflows/ci-prerelease.yml
rogercoll: Yes, it seems that is not possible to run reuse local workflows in already started jobs. Note that  `component_molecule_packaging.yml` defines the instance that it will run, which it can be different that the actual running.
The issue in this case is that we need to check out in the actual repository to get the version details. 
Initially I use the component file, but this is the error that I got: https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/4025196671/jobs/6918579337

rubenruizdegauna (approved):
LGTM

PR #14, opened 2023-02-15
NR-83106 feat: add canaries Ansible deployment

rogercoll:


test/canaries/deploy_canaries.yml
```diff
@@ -0,0 +1,39 @@
...
+  become: true
+  tasks:
+    - community.docker.docker_compose:
+        project_name: nr-otel-collector_canaries
+        definition:
+          version: '3'
+          services:
+            previous_collector:
+              container_name: "nr_otel_collector_previous"
+              image: "otel/opentelemetry-collector:{{ previous_version }}"
```
rogercoll, line 25: We need to change it once we release the corresponding docker images

test/canaries/deploy_canaries.yml
```diff
@@ -0,0 +1,39 @@
...
+  hosts: all
+  gather_facts: true
+  become: true
+  tasks:
+    - community.docker.docker_compose:
+        project_name: otel-collector_canaries
+        definition:
+          version: '3'
+          services:
+            previous_collector:
```
rubenruizdegauna, line 23: what configuration will be applied? 🤔

rogercoll: The default one: https://github.com/newrelic/opentelemetry-collector-releases/blob/main/distributions/nr-otel-collector/Dockerfile#L13

But we still don't have a configuration with NR env vars (on going https://github.com/newrelic/opentelemetry-collector-releases/pull/16)

rubenruizdegauna: 👍 I was thinking on having an external conf mounted in `/etc/nr-otel-collector/config.yaml` in docker-compose, but it makes sense, lets do it in the ongoing task

rubenruizdegauna (approved):
LGTM

test/canaries/deploy_canaries.yml
rogercoll: Great! Let's evaluate it once we have the release and a default config. I think it would be great to use the default configuration, so we test the same installation environment.

PR #23, opened 2023-03-01
NR-83103 Install requirements in pkg tests

rubenruizdegauna:


rogercoll (approved):
Great work!

rogercoll (changes requested):
We are thinking on running install-requirement on the provision step, please don't merge this yet 🙏

brushknight (approved):
LGTM

PR #26, opened 2023-03-06
NR-83112 : provision canaries

brushknight:


test/canaries/roles/otel-dockerized/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  file:
+    path: "/srv/otel_collector/"
+    state: directory
+    mode: '0755'
+- name: copy compose to the machine
+  template:
+    src: ../files/docker-compose.yml.j2
+    dest: "/srv/otel_collector/otel_collector.yml"
+    mode: 0644
+- name: set previous display_name
```
rogercoll, line 12: What do you think setting `display_name` in the docker-compose template?
Same Ansible variables can be used and we decouple env vars from the role.

test/provision/terraform/install-requirements.yml
```diff
@@ -1,6 +1,6 @@
 ---
 
-- hosts: testing_hosts_linux
+- hosts: all
```
rogercoll, line 3: ❤️

rogercoll (commented):
Nice! Left a few nits to discuss

test/canaries/roles/otel-dockerized/tasks/main.yml
brushknight: Do you have an idea how to set it? 🤔

rogercoll (approved):
LGTM

PR #34, opened 2023-04-28
NR-95227: Add third party notices

rogercoll:


Makefile.common
```diff
@@ -0,0 +1,31 @@
...
+DISTRIBUTIONS_MODS := $(shell find ./**/* $(FIND_MOD_ARGS) -exec $(TO_MOD_DIR) )
+
+## Retrive all used dependencies in the same output
+# RESULTS := $(foreach dir,$(DISTRIBUTIONS_MODS),$(shell cd $(dir) && go list -mod=mod -m -json all))
+
+
+
+.PHONY: third-party-notices
+third-party-notices: OUT ?= THIRD_PARTY_NOTICES.md
+third-party-notices:
```
DavSanchez, line 19: Could we assure here that we have the `go-licence-detector` utility so it doesn't fail if called directly? I see it's done in the `licenses` target, I guess we shouldn't be expected to run this directly, I point to this just in case it can be streamlined.

DavSanchez (dismissed):
I have a couple questions, but in general LGTM! 🔥

check.mk
```diff
@@ -0,0 +1,19 @@
...
+# Trigger each module's delegation target
+.PHONY: for-all-target
+for-all-target: $(ALL_DISTRIBUTIONS)
+
+.PHONY: licenses
+licenses: internal/tools build
+	@$(MAKE) $(FOR_GROUP_TARGET) TARGET="third-party-notices"
+
+.PHONY: licenses-check
+licenses-check: internal/tools
```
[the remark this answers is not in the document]
rogercoll: Yes, modified here https://github.com/newrelic/opentelemetry-collector-releases/pull/34/commits/0a1020965b4af2911f60e77539d9a606f458d74d to write on the same file and use git diff for check

Makefile.common
rogercoll: The Makefile.common file should only be included from inside distributions' subdirectories and is a helper in case you want to debug a target for a specific distribution because it failed from the root. The issue is that `install-tools` target can not be included in this file as it won't be detected from the distribution's file (is only available when launched from main directory).

Added a sanity check here https://github.com/newrelic/opentelemetry-collector-releases/pull/34/commits/c82062757424e4deff5015242940e430a28605dc 

Any idea on how we could call the tools makefile target?

DavSanchez: I think adding the check for the program availability with the target to run if it's not present is fine!

DavSanchez (approved):

PR #35, opened 2023-05-02
Remove not needed processors and update other components to last release (NR-109507)

alvarocabanas:
Remove not needed processors and update other components to last release


DavSanchez (approved):
lgtm

PR #46, opened 2023-05-16
NR-95230 allow to run NRDOT as root

DavSanchez:
- NR-95230 feat: run as root (systemd default)
- ci: test for process ownership by root
- ci: bump Go to v1.19.8
- feat: bump to 0.0.2
- feat: select privileged or unprivileged install
- style: enclose variable in braces
- fix: nrdot version var name
- feat: configurable installation-pinned role


rubenruizdegauna (approved):
LGTM 🔥

PR #53, opened 2023-05-19
[draft] Adds k8s attributes processor (NR-116487)

gsanchezgavier:
Adds k8s attributes processor


gsanchezgavier:
in favor of #54

PR #56, opened 2023-05-22
NR-112285 : Update default config

brushknight:


configs/nr-otel-collector-agent-linux.yaml
```diff
@@ -187,7 +198,14 @@ service:
...
-
-  extensions: [health_check]
+    traces:
+      receivers: [otlp]
+      processors: [transform, resourcedetection, batch]
+      exporters: [logging, otlp]
+    logs:
+      receivers: [otlp, filelog]
+      processors: [transform, resourcedetection, batch]
+      exporters: [logging, otlp]
```
josemore, line 209: Having the `verbosity: detailed` in the logging exporter and then keeping it here would push too much data to OTLP Logs. We need to either reduce verbosity or exclude it from Logs.

configs/nr-otel-collector-agent-linux.yaml
```diff
@@ -132,32 +142,33 @@ processors:
...
-      match_type: strict
   resource:
     attributes:
       - key: host.id
         from_attribute: host.name
         action: upsert
+
   resourcedetection:
-    detectors: ["env", "ec2", "system"]
+    detectors: ["env", "system"]
```
josemore, line 171: is `host.name` populated when the `ec2` detector is removed? If not, we'd need to keep it and add azure and gcp too.

configs/nr-otel-collector-agent-linux.yaml
rogercoll: I would vote to decrease the verbosity, what about using the default verbosity value; `normal`? https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/loggingexporter#getting-started

rogercoll (approved):
Nice!

PR #57, opened 2023-05-25
[NR-120689] disable prometheus normalization feature-gate by default

sigilioso:
It disables the `pkg.translator.prometheus.NormalizeName` feature-gate by default.

## Background information

The feature gate was enabled by default by [reaching the beta state](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/20519/files#diff-29db204fdb7feae902a4f6e8e034f2b00d5c515960dfdd124b6e85eb26d81b1eL87) in otelcollector-contrib [v0.76.3](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.76.3). When this feature-gate is enabled prometheus metrics names are renamed to match OTEL conventions [in the prometheus receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a619329714c5ca24b0e9ba2a83025c441cbf35a8/receiver/prometheusreceiver/internal/metricfamily.go#L305) and we would like to avoid this behavior by default.

Additionally, there is an ongoing discussion to consider providing a configuration option (not a feature-gate) to disable the prometheus normalization: [opentelemetry-collector-contrib#21743](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/21743).

### DIFF
--- a/distributions/nr-otel-collector/Dockerfile
+++ b/distributions/nr-otel-collector/Dockerfile
@@ -10,5 +10,5 @@ COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certifica
 COPY --chmod=755 nr-otel-collector /nr-otel-collector
 COPY configs/nr-otel-collector-agent.yaml /etc/nr-otel-collector/config.yaml
 ENTRYPOINT ["/nr-otel-collector"]
-CMD ["--config", "/etc/nr-otel-collector/config.yaml"]
+CMD ["--config", "/etc/nr-otel-collector/config.yaml", "--feature-gates=-pkg.translator.prometheus.NormalizeName"]
 EXPOSE 4317 55678 55679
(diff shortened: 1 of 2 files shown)


rogercoll (approved):
Great! Do you know if the idea (in the near future) is to remove this feature gate and be the default behavior?

sigilioso:
> Great! Do you know if the idea (in the near future) is to remove this feature gate and be the default behavior?

The feature-gate is supposed to be eventually removed, therefore the default behavior would be the same as it is at the moment: having the prometheus normalization name enabled. However, since this change is quite disruptive and affects everyone using the prometheus receiver and not using one of the prometheus exporters, including a configuration option to disable prometheus normalization is being considered [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/21743).

It seems the feature-gate will be kept for a while, but when it is removed we will need to either rely on this configuration option (if its finally implemented) or take a different approach.

rogercoll:
Sounds good! Thanks for the explanation and links @sigilioso

PR #63, opened 2023-05-31
feat(dockerStatReceiver): add docker and statsd receivers to nrdot

paologallinaharbur:
This Pr adds statsd and docker receivers to nrdot


rubenruizdegauna (approved):
LGTM

PR #74, opened 2023-07-27
NR-142172: Remove host.id rewrite

brushknight:


configs/nr-otel-collector-agent-linux.yaml
```diff
@@ -167,10 +167,6 @@ processors:
   batch:
 
   resource:
```
rogercoll, line 169:
```suggestion

```

Should we remove the resource processor from the config too?

rubenruizdegauna: good 👁️  removed in da6c864d9aa3ce59beeff9187bc004af721028ec

PR #79, opened 2023-08-07
ci: rename env provisioner module

DavSanchez:


test/provision/terraform/main.tf
```diff
@@ -29,8 +29,8 @@ variable "inventory_output" {
   default = "./inventory.ec2"
 }
 
-module "otel-env-provisioner" {
-  source             = "git::https://github.com/newrelic-experimental/otel-env-provisioner//terraform/otel-ec2"
+module "env-provisioner" {
+  source             = "git::https://github.com/newrelic-experimental/env-provisioner//terraform/otel-ec2"
```
alvarocabanas, line 33: I guess the module alerts: https://github.com/newrelic/opentelemetry-collector-releases/blob/9d674974cefe0798ecc01ee9ff2d7fd9bcd9b385/test/canaries/alerts/main.tf#L3 should also be updated, isn't it?

DavSanchez: You're totally right!

alvarocabanas: Great

alvarocabanas (approved):

PR #93, opened 2023-12-16
ci: install crowdstrike falcon

DavSanchez:


test/provision/Makefile
```diff
@@ -16,7 +16,7 @@ endif
 
 .PHONY: provision
 provision: terraform/backend
-	$(MAKE) ansible/dependencies ANSIBLE_FOLDER=$(ANSIBLE_FOLDER)
+	$(MAKE) ansible/dependencies ANSIBLE_FOLDER=$(ANSIBLE_FOLDER) ANSIBLE_ENV_SCRIPT=$(ANSIBLE_FOLDER)/set_ansible_env.sh
```
rogercoll, line 19: What do you think running the `ssh-add "$CROWDSTRIKE_ROLE_PULL_KEY"` command before doing the "make ansible/dependencies" in this Makefile so we make CROWDSTRIKE_ROLE_PULL_KEY only required in the provision environment (Makefile)?

test/Ansible.common
```diff
@@ -1,16 +1,20 @@
 ANSIBLE_FOLDER ?= $(CURDIR)
-REQUIREMENTS_FILE ?= $(ANSIBLE_FOLDER)/requirements.yml
-
-ROLES_PATH:=$(ANSIBLE_FOLDER)/roles
-COLLECTIONS_PATH:=$(ANSIBLE_FOLDER)/collections
+export REQUIREMENTS_FILE ?= $(ANSIBLE_FOLDER)/requirements.yml
+export ROLES_PATH:=$(ANSIBLE_FOLDER)/roles
+export COLLECTIONS_PATH:=$(ANSIBLE_FOLDER)/collections
+export CROWDSTRIKE_ROLE_PULL_KEY := $(HOME)/.ssh/crowdstrike_ansible_role_key
```
rogercoll, line 5: Is this variable used outside the provision directory?

test/provision/Makefile
rogercoll: Thus we could remove the ANSIBLE_ENV_SCRIPT and the variable exports.

rogercoll (commented):
Nice, added a comment related to how the Ansible Makefile is called, could we simplify it by running the ssh command in the provision Makefile?

test/Ansible.common
DavSanchez: It's used in the script along with the Ansible-related ones (roles path, collections path, reqs file...)

test/provision/Makefile
DavSanchez: I guess I can run `eval $$(ssh-agent -s) && ssh-add ...` in the same line, let me try and provision some canaries with that

rogercoll: Ah! So the issue is that the ssh-agent is not available in the Ansible dependencies shell? What about exposing a generic SSH key variable from the Ansible common Makefile and do the ssh-add if defined? I think it would be great if we only define and check the CrowdStrike variable when needed (provision)

DavSanchez: Yup, I ended up with something similar, WDYT? Let me know if you see better ways to achieve it!

test/Ansible.common
```diff
@@ -9,8 +9,18 @@ $(ROLES_PATH) $(COLLECTIONS_PATH):
 
 .PHONY: ansible/dependencies
 ansible/dependencies: $(ROLES_PATH) $(COLLECTIONS_PATH)
+ifndef CROWDSTRIKE_ROLE_PULL_KEY
 	ansible-galaxy role install -r $(REQUIREMENTS_FILE) -p $(ROLES_PATH)
 	ansible-galaxy collection install -r $(REQUIREMENTS_FILE) -p $(COLLECTIONS_PATH)
+else ifeq ($(wildcard $(CROWDSTRIKE_ROLE_PULL_KEY)),)
+	ansible-galaxy role install -r $(REQUIREMENTS_FILE) -p $(ROLES_PATH)
+	ansible-galaxy collection install -r $(REQUIREMENTS_FILE) -p $(COLLECTIONS_PATH)
```
rogercoll, line 17: Why are we doing the `ansible-galaxy` command before? Would it work to check if CROWDSTRIKE_ROLE_PULL_KEY is defined instead? If so, add the ssh key:

```
ifdef CROWDSTRIKE_ROLE_PULL_KEY
  eval $$(ssh-agent -s) && \
  ssh-add $(CROWDSTRIKE_ROLE_PULL_KEY)
endif
ansible-galaxy role install -r $(REQUIREMENTS_FILE) -p $(ROLES_PATH) && \
ansible-galaxy collection install -r $(REQUIREMENTS_FILE) -p $(COLLECTIONS_PATH)

Comment (DavSanchez) [reply test/Ansible.common:17]: I wanted to do the `ssh-add` only if the variable is defined **and** the file exists in the host (if not the recipe might fail). Not sure Makefile syntax supported intersections so I wrote it like this. Is there any way we can avoid the duplication?

Comment (rogercoll) [reply test/Ansible.common:17]: I think the Makefile should fail (ssh-add will fail) if the file defined in CROWDSTRIKE_ROLE_PULL_KEY does not exist, instead of pulling the repos without it. I would rely on the caller to provide the expected file path. Indeed, what do you think if we call it `SSH_KEY` instead of CROWDSTRIKE_ROLE_PULL_KEY in the `Ansible.common` file, we can still keep the CROWDSTRIKE keyword in the provisioning.

Comment (DavSanchez) [reply test/Ansible.common]: Not sure if we should fail if the CROWDSTRIKE_ROLE_PULL_KEY does not exist. I could be running this locally somehow, using other keys with permissions that the `ssh-agent` already has included, so I could pull the collections without this specific key. If no key is able to pull the private repo it will fail anyway, wdyt?

As for changing the var name, makes sense. Changed in https://github.com/newrelic/opentelemetry-collector-releases/pull/93/commits/291942074b2b22d7c49b96b926665e5f1886b65a!

Comment (rogercoll) [reply test/Ansible.common]: But wouldn't you have unexpected results? For the current example:
`$(MAKE) ansible/dependencies ANSIBLE_FOLDER=$(ANSIBLE_FOLDER) SSH_KEY=$(HOME)/.ssh/crowdstrike_ansible_role_key`

If `SSH_KEY=$(HOME)/.ssh/crowdstrike_ansible_role_key` does not exist, I won't be able to know I have provided an invalid path, but instead "some" roles will be fetched. I would stick with the fail-fast approach, but I might be missing something. We can discuss this async.

File: test/Ansible.common [hunk 4]
```diff
@@ -9,17 +9,14 @@ $(ROLES_PATH) $(COLLECTIONS_PATH):
...
-ifndef SSH_KEY
-	ansible-galaxy role install -r $(REQUIREMENTS_FILE) -p $(ROLES_PATH)
-	ansible-galaxy collection install -r $(REQUIREMENTS_FILE) -p $(COLLECTIONS_PATH)
-else ifeq ($(wildcard $(SSH_KEY)),)
-	ansible-galaxy role install -r $(REQUIREMENTS_FILE) -p $(ROLES_PATH)
-	ansible-galaxy collection install -r $(REQUIREMENTS_FILE) -p $(COLLECTIONS_PATH)
-else
+ifdef SSH_KEY
 	eval $$(ssh-agent -s) && \
 	ssh-add $(SSH_KEY) && \
```
rogercoll, line 14: Wdyt of setting the "endif" here, so we don't replicate the ansible-galaxy command? (add the ssh key if defined, run the ansible-galaxy command in any case)

DavSanchez: Each line of a makefile uses a sub-shell, if we do it like this we will lose the access to the ssh-agent's socket env var 😞 that's why the command inside the SSH_KEY defined clause is actually a single line

DavSanchez: Unless we somehow populate it for later use inside the same recipe I'm not sure how to achieve it

rogercoll (approved):
LGTM

test/Ansible.common
rogercoll: Make sense to me, thanks for the reply!

PR #96, opened 2024-02-23
NR-164382: unify provision strategy

rogercoll:
Update provision strategy with upstream changes: centralize AMIs


test/canaries/alerts/Makefile
```diff
@@ -42,6 +42,7 @@ ifndef TF_VAR_policies_prefix
 endif
 ifeq ("$(wildcard $(PROVISION_ALERTS_TERRAFORM_WORKSPACE)/vars.auto.tfvars)","")
 	@cp $(PROVISION_ALERTS_TERRAFORM_WORKSPACE)/vars.auto.tfvars.dist $(PROVISION_ALERTS_TERRAFORM_WORKSPACE)/vars.auto.tfvars
+	@$(PROVISION_ALERTS_TERRAFORM_WORKSPACE)/fetch_instances.sh $(TF_VAR_instance_name_pattern) >> $(PROVISION_ALERTS_TERRAFORM_WORKSPACE)/vars.auto.tfvars
```
alvarocabanas, line 45: So, the tfvars will be populated with:
```
ec2_prefix       = "TAG_OR_UNIQUE_NAME"
ec2_filters      = ["al-2023", "al-2", "debian-bookworm", "debian-bullseye", "redhat-9.0", "sles-15.4", "centos-stream", "ubuntu22.04"]
```
And then all the retrieved instances from the shell script not?

So then we use the filters to only select the ones we want from all the retrieved, is it correct?

rogercoll: Not exactly, the filters are used by the env-provisioner to provision the instances, not the alerts. The Makefile you have pointed out is used to provision and destroy the NR alerts. To create the alerts, we use the `nr-alerts` [Terraform module](https://github.com/newrelic-experimental/env-provisioner/tree/main/terraform/nr-alerts) which expects a list of current and previous display names, see: https://github.com/newrelic-experimental/env-provisioner/blob/main/terraform/nr-alerts/variables.tf#L21

The `fetch_instances.sh` scripts is used to fetch the current AWS instances (canaries) and generate the list of display names. Another option would be to hard-code them, but with this script we prevent manual updates of the list.

Let me know if you need further details or more context.

alvarocabanas: Thanks for the clear explanation, with this I have it clear.

alvarocabanas (approved):
LGTM

PR #100, opened 2024-05-09
feat: add ubuntu 24.04 packages (NR-248002)

gsanchezgavier:
add ubuntu 24.04 packages

### DIFF
--- a/.github/workflows/ci-prerelease_linux_on_demand.yml
+++ b/.github/workflows/ci-prerelease_linux_on_demand.yml
@@ -9,10 +9,13 @@ on:
       dest_prefix:
         description: 'Repo prefix'
         required: true
+      tag:
+        description: 'tag used for the testing release'
+        required: true
 
 env:
   GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  TAG: "0.2.0"
+  TAG: ${{ github.event.inputs.tag }}
   DOCKER_HUB_ID: ${{secrets.OHAI_DOCKER_HUB_ID}}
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-prerelease_linux_on_demand.yml
```diff
@@ -9,10 +9,13 @@ on:
...
         description: 'Repo prefix'
         required: true
+      tag:
+        description: 'tag used for the testing release'
+        required: true
 
 env:
   GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  TAG: "0.2.0"
+  TAG: ${{ github.event.inputs.tag }}
```
rubenruizdegauna, line 18: nice 🧹

distributions/nr-otel-collector/manifest.yaml
```diff
@@ -2,7 +2,7 @@ dist:
   module: github.com/newrelic/opentelemetry-collector-releases/nr-otel-collector
   name: nr-otel-collector
   description: New Relic OpenTelemetry Collector
-  version: 0.6.0
+  version: 0.7.0
```
rubenruizdegauna, line 5: Is this correct? We are not releasing anything new 🤔 I would wait until something is added and I would add to those release notes that the release adds support for Ubuntu 24.04.

Wdyt?

gsanchezgavier: I'm a bit afraid not to release now after some changes in the pipeline and then have to deal with some issues in the future , this is why i was planning to do a release

gsanchezgavier:
closed in favor of #108

PR #119, opened 2024-09-03
chore: increase threshold of network packet alert on canary

gsanchezgavier:
Latest release `0.8.0` triggered the network packet receive/transmit alert for some of the canaries. I didn't found any relevant data on the receiver changelog, and metrics looked to be working Ok so increasing the threshold to avoid new false positive.
image


rubenruizdegauna (approved):
LGTM, thanks for taking care of this!

PR #122, opened 2024-09-18
fix: Added endpoints to receivers

svetlanabrennan:
The nr-otel-collector only worked when I added endpoints to `grpc` and `http` for the receivers.

### DIFF
--- a/configs/nr-otel-collector-gateway.yaml
+++ b/configs/nr-otel-collector-gateway.yaml
@@ -13,7 +13,9 @@ receivers:
   otlp:
     protocols:
       grpc:
+        endpoint: 0.0.0.0:4317
       http:
+        endpoint: 0.0.0.0:4318
 
 processors:
   batch:
--- a/configs/nr-otel-collector-gateway.yaml
+++ b/configs/nr-otel-collector-gateway.yaml
@@ -13,6 +13,8 @@ receivers:
   otlp:
     protocols:
       grpc:
+        # To limit exposure to denial of service attacks, change the host in endpoints below from 0.0.0.0 to a specific network interface.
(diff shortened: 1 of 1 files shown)


configs/nr-otel-collector-gateway.yaml
```diff
@@ -13,7 +13,9 @@ receivers:
   otlp:
     protocols:
       grpc:
+        endpoint: 0.0.0.0:4317
```
DavSanchez, line 16:
```suggestion
        # To limit exposure to denial of service attacks, change the host in endpoints below from 0.0.0.0 to a specific network interface.
        # See https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md#safeguards-against-denial-of-service-attacks
        endpoint: 0.0.0.0:4317
```

DavSanchez (commented):
Could you add the comments I suggested there or at the beginning of the file? This follows the same as [what the upstream OTel Collector is doing](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/befe9509c530f7f48f5e6d7facbc27fa79ebefcb/distributions/otelcol/config.yaml#L1-L2). Thanks!

DavSanchez (approved):

kb-newrelic:
Closing this for now due to change of ownership. We're still stabilizing the project and have paused changes to the config until we have established a stable governance process. Feel free to reach out in Slack (#help-opentelemetry) if you have any questions about this.

Note: In this particular case the config is still usable as the otlp receiver is exposed under localhost, so containers running on the same node could write to it and in combination with a sidecar proxy, also containers outside of the node could. As mentioned in the [otel docs](https://opentelemetry.io/docs/collector/configuration/#basics), exposing `0.0.0.0` by default is convenient but not recommended as it greatly increases the attack surface.

PR #132, opened 2024-10-18
ci: add workflow to reset stateful test infra

kb-newrelic:
### Summary
- Add workflow to destroy packaging and canary testing infrastructure, i.e. EC2 instances launched by terraform scripts, on demand.
- Boolean flag allows destroying only canary or packaging infra. This can be helpful when canary tests were started on demand while the packaging tests were still failing and we don't want to interfere with those parallelized test strains.

### Context
- [Prerelease workflow for 0.8.3](https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/11300063884) fails repeatedly for the same reason. In order to rule out the reused EC2 instance as the culprit, we should have a workflow to reset test infrastructure on demand.
- Having this workflow allows maintainers to debug broken workflows without having to have write access to the underlying AWS test account

### DIFF
--- a/.github/workflows/clean_test_infra_on_demand.yml
+++ b/.github/workflows/clean_test_infra_on_demand.yml
@@ -1,0 +1,78 @@
+name: 💥 Clean test infra on demand
+
+on:
+  workflow_dispatch:
+    inputs:
+      unique_tag:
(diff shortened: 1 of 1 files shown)


DavSanchez (approved):
LGTM!

PR #135, opened 2024-10-24
chore: Remove all but ubuntu instances in test pipeline

mailo-nr:
We're seeing spurious failures while provisioning some of the ec2 instances, we're reducing the scope for now while we iron out the details of what os/arch we need to be testing for.


kb-newrelic:
LGTM. The main dependency for linux packages is [the superagent](https://newrelic.atlassian.net/wiki/x/AYDx4w) which just uses the [debian package](https://github.com/newrelic/newrelic-super-agent/blob/7964e8b263d490047a873b70b7d4df38df1783ae/build/embedded/embedded.yaml#L60), so we can use ubuntu as a debian derivative for testing that.

kb-newrelic (approved):

PR #143, opened 2024-11-05
fix: download trivy db only once per day

kb-newrelic:
### Summary (https://new-relic.atlassian.net/browse/NR-335800)
- Trying to address the GHCR global rate limiting issue ([that affects all users of trivy b/c its a ratelimit on the vendor namespace](https://github.com/orgs/community/discussions/139074)) by combining all of the available fixes, i.e.
    1. Use [oras](https://oras.land/) to [download trivy db only once daily and cache in workspace](https://github.com/aquasecurity/trivy-action/issues/389#issuecomment-2373539982). 
    2. Modified `oras pull ghcr.io/aquasecurity/trivy-db:2` to `{ oras pull ghcr.io/aquasecurity/trivy-db:2 } || { oras pull public.ecr.aws/aquasecurity/trivy-db:2 }` (same for java db download) to enable a fallback to ECR which [also hosts the db](https://github.com/aquasecurity/trivy-action/issues/389#issuecomment-2373539982) but can also be affected by ratelimiting.
- Changed scheduled security run to morning of PT to align with timezone of new owning team
- using ubuntu-22.04 [based on this suggestion](https://github.com/orgs/community/discussions/139074#discussioncomment-10935090). Our rationale was that it might actually pull a different artifact due to different arch/os and therefore be subject to different rate limit - might be a stretch but as there is no real solution yet, we're willing to give it a shot.
- Updating and pinning version of trivy action to newest as of today (0.28.0)

### Misc
- Make slack message more generic (less team-specific)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -64,7 +64,7 @@ jobs:
         run: echo "version=$(jq -r '.version' dist/metadata.json)" >> $GITHUB_ENV
 
       - name: Trivy security check
-        uses: aquasecurity/trivy-action@master
+        uses: aquasecurity/trivy-action@0.28.0
         with:
           image-ref: "newrelic/nr-otel-collector:${{ env.version }}-rc-amd64"
           format: 'table'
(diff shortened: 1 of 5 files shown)


mailo-nr (approved):
Looks good, hopefully we can simplify things once they address the issue on their end.

PR #144, opened 2024-11-07
Kbauer/fix syntax errors

kb-newrelic:
### Summary
- #143 is [causing issues](https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/11713709522/job/32626976898), most likely due to some syntax error
```
line 6: syntax error: unexpected end of file
```
- Fixing [potentially invalid block scalars](https://yaml.org/spec/1.2.2/#chapter-8-block-style-productions) by removing unindented newlines following script 
> It is an error if any non-[empty line](https://yaml.org/spec/1.2.2/#empty-lines) does not begin with a number of spaces greater than or equal to the content indentation level.
- According [to the docs](https://www.gnu.org/software/bash/manual/bash.html#Command-Grouping), bash requires a semicolon in `{ }` expressions. zsh is 'smart' enough to ignore that but when testing with bash it actually does cause errors or rather an input prompt, e.g. `{ echo 'foo' }` is invalid

### DIFF
--- a/.github/workflows/download_trivy_db.yml
+++ b/.github/workflows/download_trivy_db.yml
@@ -20,7 +20,7 @@ jobs:
         run: |
           mkdir -p $GITHUB_WORKSPACE/.cache/trivy/db
           # try GHCR, fallback to ECR
-          { oras pull ghcr.io/aquasecurity/trivy-db:2; } || { oras pull public.ecr.aws/aquasecurity/trivy-db:2; }
+          oras pull ghcr.io/aquasecurity/trivy-db:2 || oras pull public.ecr.aws/aquasecurity/trivy-db:2
           tar -xzf db.tar.gz -C $GITHUB_WORKSPACE/.cache/trivy/db
           rm db.tar.gz
       - name: Download and extract the Java DB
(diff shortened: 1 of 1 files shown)


.github/workflows/download_trivy_db.yml
```diff
@@ -20,19 +20,17 @@ jobs:
         run: |
           mkdir -p $GITHUB_WORKSPACE/.cache/trivy/db
           # try GHCR, fallback to ECR
-          { oras pull ghcr.io/aquasecurity/trivy-db:2 } || { oras pull public.ecr.aws/aquasecurity/trivy-db:2 }
+          { oras pull ghcr.io/aquasecurity/trivy-db:2; } || { oras pull public.ecr.aws/aquasecurity/trivy-db:2; }
```
mailo-nr, line 23: I believe we can drop the curly braces, ie: ```oras pull ghcr.io/aquasecurity/trivy-db:2 || oras pull public.ecr.aws/aquasecurity/trivy-db:2```

mailo-nr (approved):
Looks good, added a small comment but nothing blocking

.github/workflows/download_trivy_db.yml
kb-newrelic: You're right. I initially added them for clarity but given that they actually might have introduced a bug, it's probably better to lose them

mailo-nr (approved):
LGTM

PR #146, opened 2024-11-07
chore: remove snyk in favor of dependabot

kb-newrelic:
### Summary
- As discussed with [the security team](https://newrelic.slack.com/archives/C056KB65U06/p1728676617468249?thread_ts=1728675906.346859&cid=C056KB65U06), we'll rely on dependabot instead of snyk


mailo-nr (approved):
LGTM

PR #150, opened 2024-11-08
chore: simplify trivy workflow for easier debug

kb-newrelic:
### Summary
- Changing trivy workflow from split logic based on 'scheduled or not' to 'scanning latest or not'. If we're scanning latest, it can upload the results even if we run it manually. This will hopefully allow us to get to the bottom of why the [scheduled scan failed](https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/11744528669/job/32719848741) while the  [manually triggered one](https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/11728764538) succeeded.
- Removing checkout because the manual run didn't have it and we are just scanning an image that is published to docker hub and we're not using any local config file like some of the [docs](https://github.com/aquasecurity/trivy-action) suggest, so there should be no reason to checkout the repo. We might want to transition to [scanning the locally build image](https://github.com/aquasecurity/trivy-action?tab=readme-ov-file#scan-ci-pipeline) to get PR-level feedback but I would do that in a separate PR.
- Printing the sarif output to have some output to understand potential errors. We could also run trivy again with table format but I ended up going with printing the sarif file to not have to deal with discrepancies between the two runs

### DIFF
--- a/.github/workflows/component_trivy.yml
+++ b/.github/workflows/component_trivy.yml
@@ -40,7 +40,11 @@ jobs:
 
       - name: Print Trivy scan results # action can't do both table/sarif output, so we just print the sarif file
         run: |
-          cat trivy-results.sarif
+          if [[ -s trivy-results.sarif ]]; then
+            cat trivy-results.sarif
+          else
+            echo "No sarif file found"
+          fi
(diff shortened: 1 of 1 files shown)


mailo-nr (approved):
Looks good

PR #151, opened 2024-11-12
chore: print and upload trivy results also on fail

kb-newrelic:
### Summary
- In order to not skip the print/upload step on a trivy failure, we need to prepend a status check expression, otherwise it defaults to `success()` only ([see docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#failure-with-conditions))
- Added `inputs.tag == 'latest'` conditional to slack step as other tags would only be present in manual runs and we can save us the notification in that case


mailo-nr (approved):
Looks good 🤞

PR #153, opened 2024-11-13
fix: exclude <= medium severity from sarif output

kb-newrelic:
### Summary
- Turns out sarif format does not respect the `severity` parameter unless you explicitly configure it, see [update in this message](https://github.com/aquasecurity/trivy-action/issues/309#issuecomment-2057669566) and the [official input docs](https://github.com/aquasecurity/trivy-action?tab=readme-ov-file#inputs):
>limit-severities-for-sarif: By default SARIF format enforces output of all vulnerabilities regardless of configured severities. To override this behavior set this parameter to true


mailo-nr (approved):
Looks Good

PR #155, opened 2024-11-19
fix: make execute bools non-required

kb-newrelic:
### Summary
- https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/11919554058 fails due to `execute_$cmd` bools not supplied. As we already supply a default, we can just make them non-required


mailo-nr (approved):
LGTM

PR #157, opened 2024-11-25
Kbauer/simplify ci

kb-newrelic:
### Summary
Removed CI steps to reduce complexity
- `check_distro_names.sh`: Script checks distro folder names and `name` field in `manifest.yml` if they match two hardcoded values `otelcol-contrib` and `otelcol` which used to be the only otel distros. This is [no longer the case](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions), so instead of trying to maintain this script, I'd rather remove it and rely on a code review to choose an adequate name when adding a new distro which doesn't happen often.
- `ci-goreleaser.yml`: The trigger is a subset of `ci.yaml` which already runs `goreleaser` with very similar parameters. The main reason for this workflow seems to have been to make sure that goreleaser doesn't break when changes to its config or the manifest were done but I'd argue the ci workflow already covers this.
   - `ci-goreleaser` uses `--snapshot --rm-dist --skip-sign --skip-sbom --timeout 2h` and `ci` uses `--snapshot --clean --skip-sign --timeout 2h`.  `--rm-dist` was deprecated in favor of [--clean](https://goreleaser.com/deprecations/?h=rm#-rm-dist) and usage of `--skip-sbom` implies `ci` doing more than `ci-goreleaser`.


mailo-nr (approved):
Looks good

PR #158, opened 2024-11-26
Kbauer/use semver tags

kb-newrelic:
### Summary ([JIRA ticket](https://new-relic.atlassian.net/browse/NR-337872))
Moving to git tags following semantic conventions, i.e. without a distro prefix to get native support for goreleaser and similar release tools and avoid having to rely on bash scripts to bridge those gaps.
- CI previously encoded the 'distributions to release' in the prefix. Given that there is currently only one distro, we can hardcode this as input where necessary. When adding a new distro, we will publish all artifacts on the same version and increment the version based on the most impactful change to any distro (instead of maintaining a version per prefixed-version per distro). This is also what the [open-telemetry repo does](https://github.com/open-telemetry/opentelemetry-collector-releases/pull/740/files).
- Main changes are to `.github/workflows/scripts/set_version.sh` which no longer has to deal with the distro prefix (hardcoded wherever code relied upon the output of this script). Some modifications were necessary to still enforce version consistency across distros but most of the logic was borrowed from [this script checking names in the manifest](https://github.com/newrelic/opentelemetry-collector-releases/pull/157/files#diff-37cb51dfc7e46c0e73db4723c3e0ea7c9fcef08dee88080738327123ebc44af9L42-L58)

### Misc
- Removed unused `push-tags` target (copy of [releases repo](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/Makefile#L67-L74)) from `build.mk` to avoid confusion as it also relates to tagging.

### Preparation
Before this will pass CI, we'll have to merge two dummy commits with tags adhering to semantic versioning, e.g. `0.8.6` and `0.8.7`.
- Without prep, the build was failing as expected with error indicating most recent tag not adhering to semver.
- [PR for 0.8.6](https://github.com/newrelic/opentelemetry-collector-releases/pull/159)
- [PR for 0.8.7](https://github.com/newrelic/opentelemetry-collector-releases/pull/160)
- tagged both commits with annotated tags and pushed to github
```
git tag -a '0.8.6' -m 'release 0.8.6' afb1b521bbe2868336b0d384f82b4e35897f7af9
git tag -a '0.8.7' -m 'release 0.8.7' a6cc0f2a39a248e5855f9130b903517f975de8b4
git push origin 0.8.6
git push origin 0.8.7
```
- rebased this PR, [retriggering CI](https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/12125630101/job/33805999133?pr=158) with two most recent tags being semver-compliant -> CI passes

### DIFF
--- a/.github/workflows/ci-prerelease.yml
+++ b/.github/workflows/ci-prerelease.yml
@@ -88,7 +88,7 @@ jobs:
         with:
           gpg_key: 'https://download.newrelic.com/infrastructure_agent/gpg/newrelic-infra.gpg'
           repo_base_url: "http://nr-downloads-ohai-staging.s3-website-us-east-1.amazonaws.com/infrastructure_agent"
-          package_name: ${{ env.NR_DISTRO }}
+          package_name: nr-otel-collector
           package_version: ${{ env.NR_VERSION }}
           platforms: "ubuntu2204,ubuntu2404"
 
--- a/.github/workflows/ci-prerelease_linux_on_demand.yml
+++ b/.github/workflows/ci-prerelease_linux_on_demand.yml
@@ -58,10 +58,6 @@ jobs:
           ref: ${{ env.BRANCH }}
           fetch-depth: 0
 
-      - name: "Tag this commit" # required for Goreleaser
-        # force tagging to allow to have a testing release with existing tag.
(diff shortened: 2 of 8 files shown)


mailo-nr (approved):
This is very exciting.

PR #159, opened 2024-11-26
chore: dummy 0.8.6 release to move to semver tags

kb-newrelic:
Dummy release for #158  to have a dedicated dummy commit in `main` that can be tagged with tag name adhering to semver.


mailo-nr (approved):
LGTM

PR #161, opened 2024-12-03
Kbauer/use dockerized ocb

kb-newrelic:
### Summary
- Use [dockerized ocb](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder#official-release-docker-image) instead of manually compiling ocb
- Versioning
   - ocb version is implicitly defined by `otelcol_version` in manifest. This also gets rid of version in `build.mk` potentially diverging from version in manifest
   - oldest version of dockerized ocb [is 0.112.0](https://hub.docker.com/r/otel/opentelemetry-collector-builder/tags). Compiling older versions of component with it, caused all kind of weird errors, so I ended up having to bump versions to `0.112.0`
      - `>= 0.113.0` removes support for `otelcol_version` in manifest, so I decided to leave that version bump for a separate PR to not complicate things even more.
      - The only relevant change for us is `loggingexporter` getting replaced by `debugexporter`, see also this [draft PR](https://github.com/newrelic/opentelemetry-collector-releases/pull/130) we already had pending
- Refactored makefiles to
    - simplify debugging
    - avoid unnecessary builds by defining file dependencies
    - pave the way for maintaining multiple distros by using delegation targets for build (previously only for checks)

### Makefile refactoring
As part of debugging the license file generatio, I ended up refactoring and hopefully simplifying the Makefile structure as the old structure was copied from the otel repo and is more complicated than we need it to be.
- moved 'delegation targets' generated for each distro from `check.mk` to `Makefile` as they were using the variable `ALL_DISTRIBUTIONS` which was defined there and are now also used in the build targets, so it makes sense to have defined in the 'root' Makefile.
- 'build' targets were modified to be distro-specific and to not run if sources (`_build/go.mod` acting as representative) or binary are already created. They were also moved from `build.mk` to `Makefile.common` where the already distro-specific 'check' targets where defined. `build.mk` and `check.mk` ar included in `Makefile` and include top-level targets to build/check all distros.
- changed 'license' targets to depend on 'generate-sources' instead of 'build' as they only require the sources, not the binary.

### Misc
- go version is implictly defined by the ocb image, so no need to allow specifying it
- Explicitly define providers instead of relying on ocb defaults, see also [0.111.0 known bugs](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.111.0)


distributions/nr-otel-collector/THIRD_PARTY_NOTICES.md
```diff
@@ -21,166 +21,6 @@ t/.
 
 
 
-## [github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib)
```
kb-newrelic, line 24: TODO: figure out why these got deleted when running `make licenses`

kb-newrelic: - diff is due to ci generating an empty license file, so the new build process is not yet compatible with ci
- still unclear why my local machine generates a different license file than mailo-nr [on his machine](https://github.com/newrelic/opentelemetry-collector-releases/pull/130/files#diff-e4cb27b3d6a773cd0e9e99ecb1364aaa663f46303065e8e14fb0e8f198c34658) for a very similar change

kb-newrelic:
Closing this to reopen a new PR as I can't wrap my head around the license file discrepancies and want to rule out build machine reuse.

PR #163, opened 2024-12-10
Kbauer/use dockerized ocb

kb-newrelic:
### Summary
- Use [dockerized ocb](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder#official-release-docker-image) instead of manually compiling ocb
- Versioning
   - ocb version is implicitly defined by `otelcol_version` in manifest. This also gets rid of version in `build.mk` potentially diverging from version in manifest
   - oldest version of dockerized ocb [is 0.112.0](https://hub.docker.com/r/otel/opentelemetry-collector-builder/tags). Compiling older versions of component with it, caused all kind of weird errors, so I ended up having to bump versions to `0.112.0`
      - `>= 0.113.0` removes support for `otelcol_version` in manifest, so I decided to leave that version bump for a separate PR to not complicate things even more.
      - The only relevant change for us is `loggingexporter` getting replaced by `debugexporter`, see also this [draft PR](https://github.com/newrelic/opentelemetry-collector-releases/pull/130) we already had pending
- Refactored makefiles (see below)

### Makefile refactoring
As part of debugging the license file generatio, I ended up refactoring and hopefully simplifying the Makefile structure as the old structure was copied from the otel repo and is more complicated than we need it to be.
- moved 'delegation targets' generated for each distro from `check.mk` to `Makefile` as they were using the variable `ALL_DISTRIBUTIONS` which was defined there and are now also used in the build targets, so it makes sense to have defined in the 'root' Makefile.
- 'build' targets were modified to be distro-specific and to not run if sources (`_build/go.mod` acting as representative) or binary are already created. They were moved from `build.mk` to `Makefile.common` where the already distro-specific 'check' targets where defined. `build.mk` and `check.mk` are still included in `Makefile` but only include top-level targets to build/check all distros.
    - added file dependencies for build targets to avoid costly rebuilds if nothing has changed
    - moving to the delegated build process should also streamline the path to building multiple distros
- changed 'license' targets to depend on 'generate-sources' instead of 'build' as they only require the sources, not the binary.

### Misc
- go version is implictly defined by the ocb image, so no need to allow specifying it
- Explicitly define providers in manifest instead of relying on ocb defaults, see also [0.111.0 known bugs](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.111.0)


Makefile.common
```diff
@@ -1,31 +1,63 @@
...
-		-noticeTemplate $(SRC_ROOT)/internal/assets/license/THIRD_PARTY_NOTICES.md.tmpl \
-		-noticeOut $(OUT)
-	@echo '' > /tmp/tmp_notices.json
+	echo '' > $(GO_MOD_TMP_FILE);\
+		cd $(BUILD_DIR) && go list -mod=mod -m -json all >> $(GO_MOD_TMP_FILE);\
+		go-licence-detector \
+			-in $(GO_MOD_TMP_FILE) \
+			-rules $(SRC_ROOT)/internal/assets/license/rules.json \
+			-noticeTemplate $(SRC_ROOT)/internal/assets/license/THIRD_PARTY_NOTICES.md.tmpl \
+			-noticeOut $(NOTICE_OUTPUT)
```
kb-newrelic, line 31: The previous version of this caused CI to generate an empty license file. I still haven't fully wrapped my head around it but knowing that make spawns a new subshell for each command I ended up clearing the tmp file before writing to it, all in the same shell, and that did the trick.

Makefile
```diff
@@ -26,6 +26,15 @@ TO_DIR=dirname {} \; | sort | grep -E '^./'
 
 ALL_DISTRIBUTIONS := $(shell find ./distributions/* $(FIND_MOD_ARGS) -exec $(TO_DIR) )
 
+# Define a delegation target for each distribution
```
kb-newrelic, line 29: delegation targets which used to live in `check.mk`

scripts/build.sh
```diff
@@ -1,56 +1,64 @@
...
 fi
 
 if [[ "$skipcompilation" = true ]]; then
     echo "Skipping the compilation, we'll only generate the sources."
 fi
 
-echo "Distributions to build: $distributions";
+echo "Distribution to build: $distribution";
 
-for distribution in $(echo "$distributions" | tr "," "\n")
```
mailo-nr, line 35: Is the thought that we'd move back to iterating on multiple distributions once we add the next one?

mailo-nr: Oh, i see it's handled in the root Makefile 👍

mailo-nr (approved):
Looks good

PR #165, opened 2024-12-16
fix: fix CVE-2024-45337 with replace directive

kb-newrelic:
### Summary
- Fixes CVE-2024-45337, see [nightly security scan](https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/12355409541/job/34478967694)
- `x/crypto` is part of the extended stdlib and is therefore used in so many dependencies that there is no hope this will be patched in each dependency soon enough to not block our build for a long time, so introducing a `replace` directive instead


mailo-nr (approved):
LGTM

PR #166, opened 2024-12-16
chore: remove unused gnostic replace directive

kb-newrelic:
### Summary
- Directive seems to be no longer necessary. Based on the [linked issue](https://github.com/google/gnostic/issues/262), if it was still necessary, `go get -u` would fail but it didn't when executing it locally and also the dependency is no longer mentioned in the `go.mod`, so I assume it just is no longer a transitive dependency.

### TODO
- [x] merge #165 first and rebase


mailo-nr (approved):
LGTM

PR #169, opened 2024-12-18
Kbauer/fix nightly build 2

kb-newrelic:
### Summary
- Add `fetch-depth` to ensure tags are checked out as nightly otherwise creates a snapshot with version [0.0.0](https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/12395675029/job/34601944664#step:7:29) while the normal ci uses the [correct tag](https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/12362786938/job/34502821653#step:8:28)
- Only execute slow tests during nightly as fast tests already ran on merge but maintain the ability for local development to execute 'only slow tests', 'only fast tests' or all tests
   - migrated test filter mechanism to use an a string based env var to allow the above-mentioned three modes and it's also not hard to imagine that we might come up with a third test type which would be easily integrated into this pattern

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,7 +21,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          fetch-depth: 0
+          fetch-depth: 0 # required for tag metadata
 
       - name: Setup Go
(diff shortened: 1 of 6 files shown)


mailo-nr (approved):
LGTM

PR #171, opened 2024-12-19
feat: Adding provisioning of permanent EKS test cluster

mailo-nr:
Adds a permanent test cluster for deploying nightly builds.   Out of the gates we'll just use a single linux node group and eventually build out what we need longer term.  I'll have a followup pr shortly to add a simple helm release.

Should be able to pull an auth token for the cluster via:
```
data "aws_eks_cluster_auth" "this" {
  name = module.ci_e2e_cluster.cluster_name
}

### DIFF
--- a/test/terraform/modules/eks_cluster/main.tf
+++ b/test/terraform/modules/eks_cluster/main.tf
@@ -1,0 +1,39 @@
+locals {
+  iam_role_permissions_boundary = "arn:aws:iam::${var.account_id}:policy/resource-provisioner-boundary"
+}
+
+module "eks_cluster" {
+  source  = "terraform-aws-modules/eks/aws"
+  version = "20.31.4"
+
+  cluster_name    = var.name
+  cluster_version = "1.31"
+
+  cluster_endpoint_private_access = true
+  cluster_endpoint_public_access  = true
+
+  vpc_id     = var.vpc_id
+  subnet_ids = var.subnet_ids
+
+  enable_cluster_creator_admin_permissions = true
+  iam_role_permissions_boundary            = local.iam_role_permissions_boundary
+
+  eks_managed_node_group_defaults = {
+    iam_role_permissions_boundary = local.iam_role_permissions_boundary
+
+    instance_types = ["m5.large"]
+
+    min_size     = 2
+    max_size     = 2
+    desired_size = 2
+  }
+
+  eks_managed_node_groups = {
+    default = {
+      # Starting on 1.30, AL2023 is the default AMI type for EKS managed node groups
+      ami_type = "AL2023_x86_64_STANDARD"
+    }
+  }
+}
+
+
--- a/test/terraform/modules/eks_cluster/main.tf
+++ b/test/terraform/modules/eks_cluster/main.tf
@@ -1,6 +1,53 @@
 locals {
   iam_role_permissions_boundary = "arn:aws:iam::${var.account_id}:policy/resource-provisioner-boundary"
+
+  # We need to ensure we don't attempt to use any of the disallowed zones:
+  # https://docs.aws.amazon.com/eks/latest/userguide/network-reqs.html
+  valid_availability_zones = [
+    for zone in data.aws_availability_zones.available.names : zone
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: test/terraform/permanent/main.tf [hunk 1]
```diff
@@ -1,13 +1,25 @@
...
+    values = ["us-east-1a", "us-east-1b"]
   }
-}
\ No newline at end of file
+}
+
+module "ci_e2e_cluster" {
+  source = "../modules/eks_cluster"
+
+  name       = "ci-e2e"
```
mailo-nr, line 20: This could be changed to `ci-e2etest` to match the nightly job, open to other ideas as well

kb-newrelic: I think keeping it separate is a good idea so that we can also distinguish it easily in NR1. Might even be good to include `aws` or in the name to make that even more explicit

test/terraform/modules/eks_cluster/main.tf
```diff
@@ -0,0 +1,39 @@
...
+
+module "eks_cluster" {
+  source  = "terraform-aws-modules/eks/aws"
+  version = "20.31.4"
+
+  cluster_name    = var.name
+  cluster_version = "1.31"
+
+  cluster_endpoint_private_access = true
+  cluster_endpoint_public_access  = true
```
kb-newrelic, line 13: Given that this cluster will run permanently, what's our strategy to protect ourself against abuse from the random internet? Not entirely sure how auth works in this case.

kb-newrelic: Looks like [it's secured via IAM](https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html), so it would be good to understand where the role requirements are configured.

mailo-nr: Good call out, was thinking we'd add the gha cidr blocks eventually: https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html#modify-endpoint-access

test/terraform/permanent/main.tf
```diff
@@ -1,13 +1,25 @@
...
+  filter {
+    name = "vpc-id"
+    values = [data.aws_vpc.this.id]
+  }
 
-  tags = {
-    created_by = "terraform"
+  filter {
+    name = "availability-zone"
+    values = ["us-east-1a", "us-east-1b"]
```
kb-newrelic, line 13: As far as I can tell these subnets are public (ingress and egress) while we probably only need egress. I'm wondering if we should create private subnets (no internet ingress b/c all telemetry generating apps will run within the cluster) and route internet egress (to NR) through a NAT gateway. We could also separate this out into a dedicated ticket. I admit that my network knowledge/skills are limited, so I might be oversimplifying things/effort.

test/terraform/modules/eks_cluster/main.tf
kb-newrelic: Interesting idea, maybe we can put it into a dedicated ticket to not lose track of it

mailo-nr: For IAM, since we're enabling `enable_cluster_creator_admin_permissions` the resource provisioner gets admin access to the cluster, we don't have access unless we add another access entry by default.

test/terraform/permanent/main.tf
mailo-nr: Good thought, I we may need to look at what level of connectivity we'll need with GH in order to orchestrate the cluster.

test/terraform/modules/eks_cluster/main.tf
kb-newrelic: Given that we have access to the credentials for that user, it should be enough to have that configured I guess. LGTM then

test/terraform/permanent/main.tf
mailo-nr: Looks like the recommendation is private, i assume the eks endpoint is separate so the above comment isn't relevant.  I'll add a couple private subnets in different zones.

kb-newrelic (approved):
LGTM

PR #174, opened 2024-12-20
feat: Deploy permanent backend into test cluster

mailo-nr:
Deploys our test helm-chart into the permanent eks cluster.  Currently it will just pull the most latest image until we add a test repository and/or publish nightly builds.


kb-newrelic (approved):
LGTM

PR #178, opened 2025-01-08
Kbauer/add tf based bare metal test

kb-newrelic:
### Summary
- Add EC2 instance running ubuntu 24.04 to nightly test infra. Using latest collector until we have a nightly version
- Add NRQL-based tests that assert that nightly test infra produces expected hostmetrics
  - `system.paging.usage`: requires a swapfile to be allocated, [added](https://repost.aws/knowledge-center/ec2-memory-swap-file) it for EC2 but skip this test for k8s as the overhead of creating an AMI for k8s nodes to use seems to much to support this test case.
  - `system.cpu.load_average.15m`: [drops to 0](https://onenr.io/0LwGl9qm8R6) in idle for the EC2, so after running long enough this test would fail, thus I changed it to `>=0`. Alternative would be to add a load-generating script.

#### TODO
- [ ] add EC2 for ubuntu22 to test on two different os versions

#### Hostname changes
- Previous PR added `OTEL_RESOURCE_ATTRIBUTES` specifier to force predictable hostnames to be reported to NR. Added some structure to those names to enforce consistency across different test environments (local-dev1, local-dev2, ci, nightly) and host machines (k8s-node1, k8s-node2, ec2-ubuntu,...) 
  - Appended the node name to k8s nodes as the daemonset deploys the collector to multiple/different nodes, so using the same hostname doesn't make sense and also causes issues in the UI because two entities with the same hostname got created

#### Removal of TestMain
- No longer use `TestMain` for test setup as test tagging/filtering should be the first logic to run to avoid errors due to missing env vars that are not necessary for the subsets of tests run, e.g. NR_INGEST_KEY not required for nightly tests. `testing.M` doesn't provide an API to skip a test, so it can't be used in the `testutil` package.

### Test
- Nightly tests specified in this PR don't pass yet because a) ec2 is not deployed yet and b) the eks node collector still uses the old host name. To ensure the tests do work, I temporarily deployed the EC2 for a few hours and ran the tests using the old k8s hostname using the following patch:
[nightly-test.patch](https://github.com/user-attachments/files/18351771/nightly-test.patch)

### DIFF
--- a/.github/workflows/ci_nightly.yaml
+++ b/.github/workflows/ci_nightly.yaml
@@ -55,7 +55,7 @@ jobs:
           cluster_name: ${{ env.TEST_CLUSTER_NAME }}
           wait: 60s
 
-      - name: Run local e2e tests
+      - name: Run slow local tests
         run: |
           IMAGE_TAG=${{ env.version }}-rc-amd64 \
           KIND_CLUSTER_NAME=${{ env.TEST_CLUSTER_NAME }} \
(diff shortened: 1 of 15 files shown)


kb-newrelic:
Closing this for now to pull nightly build via private S3 bucket

PR #192, opened 2025-01-22
ci: enforce node-naming conventions

kb-newrelic:
### Summary
- Use `nightly` host name prefix for EKS nodes. Still using `permanent` which causes [nightly tests to fail](https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/12914171239/job/36013128738#step:14:43).
- Replace `k8s-node` with `k8s_node` for consistency in regular ci tests
- [Current host names](https://onenr.io/0OwvdMG2yjv) reported to NR for reference
image


mailo-nr (approved):
LGTM

PR #193, opened 2025-01-22
ci: remove ansible-based tests

kb-newrelic:
### Summary
- Nightly tests running hostmetrics tests on EKS (run as container) and EC2 (install + run on vm) were introduced which should cover the most crucial part of the existing ansible tests which we want to remove to keep the maintenance burden manageable.
- keep `test/terraform/fargate` and `test/terraform/Aws.common` in place until we have coordinated the removal of resources defined in there


mailo-nr (approved):
🧨

PR #194, opened 2025-01-22
feat: Move to multiple distro and simplify ci

mailo-nr:
Large refactor that will facilitate multi-distribution publishing.

Main changes are as follows:

- Simplifies the build process by utilizing existing tools and relies on act as a task runner for working locally.  There is how a Brewfile that will install local deps using `brew bundle install` as well as a placeholder secrets file.
- Updates otel components to 1.18.0
- Adds a cli tool to generate goreleaser configs (default and nightly).  This will help keep configs clean and prevent accidentally pushing changes that would modify the build.


.github/workflows/ci.yaml
```diff
@@ -29,25 +34,55 @@ jobs:
...
+        run: make ci DISTRIBUTIONS=${{ matrix.dist }}
 
-      - name: Validate distributions third party libraries
-        run: make licenses-check
+      - name: Login to Docker
+        uses: docker/login-action@v3
+        with:
+          registry: docker.io
+          username: ${{ secrets.OTELCOMM_DOCKERHUB_USERNAME }}
+          password: ${{ secrets.OTELCOMM_DOCKERHUB_PASSWORD }}
```
kb-newrelic, line 48: Do we need this step? My naive thinking is that this workflow would never need to log into docker as it won't publish anything and the built image will only be used locally. Am I missing something?

distributions/nr-otel-collector/.goreleaser.yaml
```diff
@@ -0,0 +1,146 @@
...
+  split: true
+dockers:
+  - goos: linux
+    goarch: amd64
+    dockerfile: Dockerfile
+    image_templates:
+      - newrelic/nr-otel-collector:{{ .Version }}-amd64
+      - newrelic/nr-otel-collector:latest-amd64
+    extra_files:
+      - config.yaml
```
kb-newrelic, line 81: Unless I am missing something, this doesn't align with the current naming and location of the configs? Either the default config would need to be moved up one level to the distro folder or this references the specific config which will be the default

.github/workflows/ci.yaml
mailo-nr: Unfortunately we do in order to work locally.  I generated read only tokens to prevent this from being an issue locally but we can also gate it via: https://nektosact.com/usage/index.html#skipping-steps

distributions/nrdot-collector-k8s/manifest.yaml
```diff
@@ -0,0 +1,53 @@
...
+receivers:
+  # shared
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver v0.118.0
+  # daemonset-only
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/filelogreceiver v0.118.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/hostmetricsreceiver v0.118.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kubeletstatsreceiver v0.118.0
+  # deployment-only
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/k8seventsreceiver v0.118.0
+  - gomod: go.opentelemetry.io/collector/receiver/otlpreceiver v0.118.0 # only in commented out gateway pipeline but useful for piping
```
kb-newrelic, line 17: would probably remove comments like this one and `# added`

.github/workflows/ci.yaml
```diff
@@ -29,25 +34,55 @@ jobs:
           go-version: '1.23'
           check-latest: true
 
+      - name: Tidy go.mod files
+        run: go mod tidy
+
```
kb-newrelic, line 39: Should about error out if there is a diff?

.github/workflows/ci.yaml
```diff
@@ -56,28 +91,31 @@ jobs:
           cluster_name: ${{ env.TEST_CLUSTER_NAME }}
           wait: 60s
 
+      - uses: azure/setup-helm@v4.2.0
```
kb-newrelic, line 94: I assume this is necessary because `test/e2e/Makefile` [requires helm for any target](https://github.com/newrelic/opentelemetry-collector-releases/blob/a85d6f703d56b38a8b2bf63ab9de7c61c9432631/test/e2e/Makefile#L12-L14) even though it is only used for `local_helm-cleanup`. I'd say we should remove `helm` from that list instead to avoid having unnecessary setup actions like this one.

.github/workflows/ci.yaml
```diff
@@ -56,28 +91,31 @@ jobs:
...
-          TRIVY_SKIP_DB_UPDATE: true
+          # dbs are downloaded async in trivy-cache.yml
+          TRIVY_SKIP_DB_UPDATE: ${{ !env.ACT }}
           TRIVY_SKIP_JAVA_DB_UPDATE: true
+
   terraform:
-    uses: ./.github/workflows/component_terraform.yml
-    if: github.event.pull_request.user.login != 'dependabot[bot]'
+    uses: ./.github/workflows/terraform.yml
+    if: ${{ !github.event.act && github.event.pull_request.user.login != 'dependabot[bot]' }}
```
kb-newrelic, line 118: I'd suggest we check for act in a uniform way unless there is a semantic difference between `github.event.act` and `env.ACT`?

Brewfile
```diff
@@ -0,0 +1,6 @@
+brew 'go@1.23'
+brew 'goreleaser/tap/goreleaser'
+brew 'gnupg'
+brew 'pinentry-mac'
```
kb-newrelic, line 4: Out of curiosity, what is this used for?

Makefile
```diff
@@ -1,45 +1,106 @@
...
-# Define the list of subdirectories that contain a Makefile
-SUBDIRS := $(patsubst ./%/Makefile,%,$(shell find $(FIND_PATH) -name Makefile))
-TARGETS := $(SUBDIRS)
+GO ?= go
+GORELEASER ?= goreleaser
 
-.PHONY: all $(TARGETS) clean $(addsuffix -clean,$(TARGETS)) help
+# SRC_ROOT is the top of the source tree.
+SRC_ROOT := $(shell git rev-parse --show-toplevel)
+OTELCOL_BUILDER_VERSION ?= 0.118.0
```
kb-newrelic, line 6: Might be worth writing a small unit test to keep this in sync with the component versions

Brewfile
mailo-nr: It's allows storing the passphrase in the osx keychain, i'm fine removing it

Makefile
mailo-nr: Definitely, i'm actually working on a bash script to catch this.

distributions/nrdot-collector-k8s/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+USER ${USER_UID}
+
+COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --chmod=755 nrdot-collector-k8s /nrdot-collector-k8s
+ENTRYPOINT ["/nrdot-collector-k8s"]
+# `4137` and `4318`: OTLP
+# `55678`: OpenCensus
+# `55679`: zpages
+# `6831`, `14268`, and `14250`: Jaeger
+# `9411`: Zipkin
```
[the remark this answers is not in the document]
mailo-nr: Yeah, this was a copypasta, I can remove those unused ports

.github/workflows/ci.yaml
kb-newrelic: Interesting, thanks for clarifying!

Brewfile
kb-newrelic: I'll for sure play around with it but might be something we want to leave out of VCS. I'll leave it up to you.

distributions/nrdot-collector-k8s/test-spec.yaml
```diff
@@ -0,0 +1,3 @@
+nightly:
+  ec2:
+    enabled: true
```
kb-newrelic, line 3: should be disabled as EC2 tests would require binaries which are not built for k8s

Makefile
kb-newrelic: Will be solved in different PR

kb-newrelic (approved):

PR #198, opened 2025-01-24
fix: skip nightly EC2 tests for paging.usage

kb-newrelic:
### Summary
- Disable nightly tests for `system.paging.usage` for EC2s until we have a better understanding on why the metrics stopped reporting. Most likely a test setup issue.


mailo-nr (approved):
LGTM

PR #200, opened 2025-01-24
ci: support distro-dependent tests

kb-newrelic:
### Summary
- Generalize fast and slow tests to support running with a second distro published under the name `newrelic/$distro_name`.
- Nightly will be adjusted in a separate PR which will also generalize the hostname to incorporate the distro to avoid unintended overlap in queries. Post this PR this is technically the case for `slow` tests but as we still have only one distro, this is okay as a temporary state and can be addressed in the follow-up PR.

### Tests
- Tested by locally modifying `.goreleaser.yaml` to publish docker images to `newrelic/foobar` and running makefile with`DISTRO=foobar`.
- Running fast tests for two distros at the same time was successful 
image

### DIFF
--- a/test/charts/mocked_backend/Chart.yaml
+++ b/test/charts/mocked_backend/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
 name: nrdot-mocked-backend
-description: A Helm chart for deploying nr-otel-collector as a DaemonSet writing to a Mock Backend for assertions in tests
+description: A Helm chart for deploying an nrdot collector as a DaemonSet writing to a Mock Backend for assertions in tests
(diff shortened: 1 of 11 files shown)


test/charts/nr_backend/values.yaml
```diff
@@ -1,12 +1,11 @@
-image:
-  repository: newrelic/nr-otel-collector
-  tag: latest
-  # Avoid accidentally pulling remote images in CI
-  pullPolicy: Never
-
```
kb-newrelic, line 6: Breaking change for terraform, still need to adjust helm chart variables there

mailo-nr (approved):
LGTM

PR #201, opened 2025-01-27
ci: support skip nightly tests by distro/host

kb-newrelic:
### Summary
- Introduce per-distro `test-spec.yaml` to toggle nightly tests. The idea was that this could be a starting point to also incorporate features like skipping specific metrics and/or listing the metrics to test in the first place.
   - Corresponding nightly test infra is only deployed when enabled, e.g. `k8s` distro will not need to run on bare EC2s, so the EC2s shouldn't be deployed in the first place.
   - Skipped toggling k8s_node tests for now as both `host` and `k8s` will need to support it and when we move to testing `k8s` in host/gateway mode, we'll need to provide more finegrained control at which point we can incorporate skipping parts of it for the `host` distro
- Incorporate distro into
    - k8s namespace to minimize interference
    - the hostname to 'namespace' data in NR

### Tests
- [tf plan](https://github.com/newrelic/opentelemetry-collector-releases/actions/runs/12997624550/job/36249015987?pr=201#step:5:1237) looks good

### DIFF
--- a/test/e2e/util/chart/mocked_backend.go
+++ b/test/e2e/util/chart/mocked_backend.go
@@ -27,8 +27,3 @@ func (m *MockedBackendChart) RequiredChartValues() map[string]string {
 		"image.tag":        envutil.GetImageTag(),
 	}
 }
-
-func (m *MockedBackendChart) GetCollectorSelector() string {
-	//TODO implement me
-	panic("implement me")
-}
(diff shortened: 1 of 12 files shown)


test/e2e/util/chart/mocked_backend.go
```diff
@@ -27,8 +27,3 @@ func (m *MockedBackendChart) RequiredChartValues() map[string]string {
 		"image.tag":        envutil.GetImageTag(),
 	}
 }
-
-func (m *MockedBackendChart) GetCollectorSelector() string {
-	//TODO implement me
-	panic("implement me")
-}
```
kb-newrelic, line 34: Unused

mailo-nr (approved):
LGTM

PR #202, opened 2025-01-28
feat: add nrdot-collector-host distro

kb-newrelic:
### Summary
- Added host distro named `nrdot-collector-host`. Files are copied from `nr-otel-collector` and modified as necessary:
  - Removed feature gate (`--feature-gates=-pkg.translator.prometheus.NormalizeName`) from default args (Dockerfile and `.conf` file) as distro does not contain [prometheus-related components](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/translator/prometheus/README.md#full-normalization)
  - Removed ports `55678` and `55679` from docker `EXPOSE` as they are related to [OpenCensus](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/opencensusreceiver#getting-started) which we do not use
  - Bumped provider versions to `1.24.0` to match component version of `0.118.0` (see [otel release notes](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.118.0))
    - Added missing providers to k8s manifest
- Rearranged/added README content to support multiple distros

### DIFF
--- a/distributions/nrdot-collector-host/Dockerfile
+++ b/distributions/nrdot-collector-host/Dockerfile
@@ -1,9 +1,7 @@
-FROM alpine:3.18 as certs
+FROM alpine:3.21 as certs
 RUN apk --update add ca-certificates
 
-FROM alpine:3.18
-
-RUN apk add --no-cache bash
+FROM scratch
 
 ARG USER_UID=10001
 USER ${USER_UID}
--- a/distributions/nrdot-collector-host/Dockerfile
+++ b/distributions/nrdot-collector-host/Dockerfile
@@ -1,0 +1,17 @@
+FROM alpine:3.18 as certs
+RUN apk --update add ca-certificates
+
+FROM alpine:3.18
+
+RUN apk add --no-cache bash
+
+ARG USER_UID=10001
+USER ${USER_UID}
+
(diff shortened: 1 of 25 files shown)


test/terraform/permanent/main.tf
```diff
@@ -1,6 +1,6 @@
 locals {
   distros = toset(distinct(flatten([
-    for _, v in fileset(path.module, "../../../distributions/**") :
+    for _, v in fileset(path.module, "../../../distributions/*/**") :
```
kb-newrelic, line 3: Exclude README from `fileset` to avoid regex failing.

distributions/nrdot-collector-host/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM alpine:3.18 as certs
+RUN apk --update add ca-certificates
+
+FROM alpine:3.18
```
mailo-nr, line 4: Any concern with us upgrading to `3.21`?

kb-newrelic: Good call, I ended up replacing it with `scratch` though and removed installing of bash. Otel does the same and no real reason to keep either.

mailo-nr (approved):
LGTM

PR #204, opened 2025-01-28
ci: fix EC2 install script

kb-newrelic:
### Summary
- Downloading `deb` package from S3 on EC2s fails as `latest_deb_package_filename` expression also matches the checksum files and files from older versions. Adding a `$` anchor filters out the checksum/signature files and `head` filters out older versions.
- Also added a grep for the collector distro as that will also result in non-unique results


mailo-nr (approved):
LGTM

PR #205, opened 2025-01-29
ci: revert normalizing host.name in query

kb-newrelic:
### Summary
- Reverting the 'normalization' of the distro in the host name as it causes nightly to fail. I initially thought it would be worth for readability but it also forces normalizing it in terraform which just becomes cumbersome to remember. If readability becomes an issue, adopting a different separator than `-` might be preferable than trying to shoehorn all segments that naturally contain hyphens into using a different character.


mailo-nr (approved):
LGTM

PR #208, opened 2025-02-04
feat: Use shared workflow for ci and nightly

mailo-nr:
This pr updates the nightly job to account for the multiple distro pattern we've adopted.

Here's an overview of the changes:

- Tweaks to the goreleaser generator to exclude `cosign` specific configs
- Creates a `ci-base` shared workflow to make things a bit more DRY
- Renames our workflows to use the recommended `.yaml` extension
- Exposes `IMAGE_REPO` in our tests to account for ECR images


cmd/goreleaser/internal/configure.go
```diff
@@ -393,9 +396,8 @@ func archName(arch, armVersion string) string {
 func Sign() []config.Sign {
 	return []config.Sign{
 		{
-			Artifacts:   "all",
-			Signature:   "${artifact}.sig",
-			Certificate: "${artifact}.pem",
```
mailo-nr, line 398: Since we're not using cosign we can remove these for now, same for the above change.

cmd/goreleaser/internal/configure.go
```diff
@@ -71,13 +71,16 @@ func Generate(dist string, nightly bool) config.Project {
...
-		DockerSigns:     DockerSigns(),
 		Version:         2,
 		Changelog:       config.Changelog{Disable: "true"},
 		Snapshot: config.Snapshot{
 			VersionTemplate: "{{ incpatch .Version }}-SNAPSHOT-{{.ShortCommit}}",
 		},
 		Blobs: Blobs(dist, nightly),
+		Release: config.Release{
+			// Disable releases on all distros for now
+			Disable: "true",
```
mailo-nr, line 82: We can update this once we're ready, no need to accidentally publish something until we're ready.

.github/workflows/ci-nightly.yaml
```diff
@@ -12,18 +12,44 @@ on:
...
+      gpg_private_key: ${{ secrets.OHAI_GPG_PRIVATE_KEY_BASE64 }}
+      gpg_passphrase: ${{ secrets.OHAI_GPG_PASSPHRASE }}
+      registry: '${{ secrets.OTELCOMM_AWS_TEST_ACC_ACCOUNT_ID }}.dkr.ecr.us-east-1.amazonaws.com'
+      nr_backend_url: ${{ secrets.NR_STAGING_BACKEND_URL }}
+      nr_ingest_key: ${{ secrets.OTELCOMM_NR_INGEST_KEY }}
+      nr_account_id: ${{ secrets.OTELCOMM_NR_TEST_ACCOUNT_ID }}
+      nr_api_base_url: ${{ secrets.NR_STAGING_API_BASE_URL }}
+      nr_api_key: ${{ secrets.OTELCOMM_NR_API_KEY }}
+
+  publish:
```
mailo-nr, line 41: I think we should eventually switch to using `artifact-upload`/`artifact-download` so we don't need to rebuild here after local testing.  Ideally with goreleaser pro we can use the `continue` functionality of goreleaser pro.

.github/workflows/ci-base.yaml
```diff
@@ -0,0 +1,169 @@
...
+          DISTRO: ${{ inputs.distribution }}
+          NR_BACKEND_URL: ${{ secrets.nr_backend_url }}
+          NR_INGEST_KEY: ${{ secrets.nr_ingest_key }}
+          NR_API_KEY: ${{ secrets.nr_api_key }}
+          NR_ACCOUNT_ID: ${{ secrets.nr_account_id }}
+          NR_API_BASE_URL: ${{ secrets.nr_api_base_url }}
+        run: |
+          make -f ./test/e2e/Makefile ci_test-slow
+
+      - name: Trivy security check
```
kb-newrelic, line 157: We didn't use to run trivy during nightly I think. We can run it there as well, no big deal but wanted to make sure this was an intentional choice

.github/workflows/ci-nightly.yaml
```diff
@@ -12,18 +12,44 @@ on:
...
 
 jobs:
-  slow-tests:
-    name: Run slow tests
+  build:
+    name: Build & Validate
+    strategy:
+      matrix:
+        distribution:
+          - nr-otel-collector
```
kb-newrelic, line 23: Missing `nrdot-collector-host`

.github/workflows/ci-base.yaml
mailo-nr: It wasn't intentional, i'll turn it off, it's slow enough already and we scan independently already

kb-newrelic (approved):

PR #213, opened 2025-02-07
fix: add missing distro folder/namespace to s3 path

kb-newrelic:
### Summary
- Incorporate [missing distro name](https://github.com/newrelic/opentelemetry-collector-releases/blob/50b9f03b0430803fce4e726a9463dc6d2d8c2f46/distributions/nrdot-collector-host/.goreleaser-nightly.yaml#L121) into s3 path when querying for binaries in EC2 on startup


mailo-nr (approved):
LGTM

PR #215, opened 2025-02-11
Kbauer/add config to k8s distro

kb-newrelic:


distributions/nrdot-collector-k8s/sync-configs.sh
```diff
@@ -0,0 +1,51 @@
...
+    } | {
+      # expose ingest endpoint via env var to align with other distros
+      sed 's/"https:\/\/otlp.nr-data.net"/${env:OTEL_EXPORTER_OTLP_ENDPOINT:-https:\/\/otlp.nr-data.net}/g'
+    } | {
+      # normalize env var names
+      sed 's/env:NR_LICENSE_KEY/env:NEW_RELIC_LICENSE_KEY/g'
+    } | {
+      # add healthcheck
+      yq '. + {"extensions": {"health_check":{}}}' |
+      yq '.service += {"extensions": ["health_check"]}'
```
kb-newrelic, line 41: Resulting yaml syntax is slightly different (`extensions: - health_check` vs `extensions: [health_check]` in our other distros) but equivalent as far as I understand, and I couldn't change it with [yq' style option](https://mikefarah.gitbook.io/yq/operators/style) as far as I can tell, so accepting it for now.

Makefile
```diff
@@ -104,3 +104,18 @@ licenses-check: licenses
 			exit 1;\
 		} \
 		|| exit 0
+
+.PHONY: sync-k8s-configs
```
mailo-nr, line 108: What are you thoughts on putting this in a Makefile inside that distributions folder since this functionality is truly unique to that distro?

kb-newrelic: Done

mailo-nr (approved):
Looks good

PR #217, opened 2025-02-14
chore: remove legacy distro nr-otel-collector

kb-newrelic:
### Summary
- nrdot-collector-host and nrdot-collector-k8s replace the existing nr-otel-collector for all known use cases, so we're removing it to reduce build times.


mailo-nr (approved):
LGTM, excited for the 🚤

PR #218, opened 2025-02-14
Kbauer/apply helm and ec2 during nightly

kb-newrelic:
### Summary
- Apply helm and EC2 during nightly instead of as part of permanent
- Use workspaces to namespace the nightly terraform state by distro, see also [terraform s3 docs](https://developer.hashicorp.com/terraform/language/backend/s3#state-storage)
- removed terraform-debug workflow as we shouldn't need it anymore with act
- [x] should be merged after #215


.github/workflows/ci-nightly.yaml
```diff
@@ -121,7 +123,50 @@ jobs:
           version: '~> v2'
           args: --skip=announce,validate --clean --timeout 2h --config .goreleaser-nightly.yaml
           workdir: distributions/${{ matrix.distribution }}
+      - name: Extract Docker Manifest SHA
+        id: extract_docker_manifest_sha
+        run: |
+          echo "docker_manifest_sha=$(echo '${{ steps.goreleaser.outputs.artifacts }}' |
```
kb-newrelic, line 129: docker manifest sha is only present here and not in `ci-base.yaml`. Looks like the manifest only gets created during publish

mailo-nr (approved):
LGTM

PR #219, opened 2025-02-18
docs: document difference between helm and distro

kb-newrelic:
### Summary
- Add some more user-friendly documentation on differences between distro and helm chart for k8s distro


mailo-nr (approved):
LGTM

PR #222, opened 2025-02-19
fix: relax condition on steal metric

kb-newrelic:
### Summary
- Nightly [fails](https://github.com/newrelic/nrdot-collector-releases/actions/runs/13401218903/job/37432282013#step:17:92) due to new requirement of cpu util for `state=steal` but it's [just 0](https://onenr.io/0PwJ943M5Q7) which is fine as far as I understand the metric
image


mailo-nr (approved):
LGTM

PR #223, opened 2025-02-19
feat: provide env var defaults for docker

kb-newrelic:
### Summary
- Supply env var defaults via config instead of systemd config to also provide defaults when used as docker image
- Dropped the port for simplicity as 443 supports both gRPC and http, see [NR docs](https://docs.newrelic.com/docs/opentelemetry/best-practices/opentelemetry-otlp/#note-ports). As far as I can tell, the old systemd default port was also wrong as it was for gRPC while the config used the otlphttpexporter


mailo-nr (approved):
LGTM

PR #224, opened 2025-02-19
fix: namespace kind clusters used for slow tests

kb-newrelic:
### Summary
- Namespacing clusters used for nightly slow tests to avoid interference between tests of different distros


mailo-nr (approved):
LGTM

PR #225, opened 2025-02-19
fix: override namespace of clusterrole

kb-newrelic:
### Summary
- the slow test didn't supply a namespace value, leaving the ClusterRole with the (wrong) default namespace name, leading to [permissions issues for nightly](https://github.com/newrelic/nrdot-collector-releases/actions/runs/13403283329/job/37438427097#step:17:444)


mailo-nr (approved):
LGTM

PR #228, opened 2025-02-19
fix: retry NR backend validations

kb-newrelic:
### Summary
- Nightly run [failed due to missing metric](https://github.com/newrelic/nrdot-collector-releases/actions/runs/13419200827/job/37487529735#step:17:105) but metric [did get ingested](https://onenr.io/02wdo549GQE), so I assume it's a race condition. Adding some retry behavior to not immediately fail but account for variable ingest delay


mailo-nr (approved):
LGTM

PR #229, opened 2025-02-19
fix: avoid use of index when reading ecr

kb-newrelic:
### Summary
- fix [remnant](https://github.com/newrelic/nrdot-collector-releases/actions/runs/13419200827/job/37493600871#step:6:102) of refactoring that happened during #218 - tf used to handle all distros instead of namespacing it via workspaces. ecr data construct now uniquely identifies a distro's repo, so no need for indexing.


mailo-nr (approved):
LGTM

PR #231, opened 2025-02-19
fix: namespace cluster roles for multi chart install

kb-newrelic:
### Summary
- tf nightly apply failed due to the [non-namespaced](https://kubernetes.io/docs/reference/kubernetes-api/authorization-resources/cluster-role-v1/) `ClusterRole` running into conflicts.
- Adding `meta.namespace` wouldn't have helped if I understand [the docs](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/#ObjectMeta) correctly as `ClusterRole` falls into
> Not all objects are required to be scoped to a namespace - the value of this field for those objects will be empty.


mailo-nr (approved):
🚀

PR #236, opened 2025-02-20
fix: properly retry assertions

kb-newrelic:
### Summary
- The old logic was flawed (didn't retry the actual request and didn't sleep), just re-checked the response

### DIFF
--- a/test/e2e/util/assert/nr_metric_assertion.go
+++ b/test/e2e/util/assert/nr_metric_assertion.go
@@ -63,26 +63,28 @@ type NrAssertion struct {
 
 func (m *NrMetricAssertion) ExecuteWithRetries(t testing.TB, client *newrelic.NewRelic, retries int, retryInterval time.Duration) {
 	query := nrdb.NRQL(m.AsQuery())
-	response, err := client.Nrdb.Query(envutil.GetNrAccountId(), query)
-	if err != nil {
-		t.Fatal(err)
-	}
-	if len(response.Results) != len(m.Assertions) {
-		t.Fatalf("Query results (%+v) and number of assertions (%+v) do not match", response.Results, m.Assertions)
-	}
-	for i, assertion := range m.Assertions {
-		for attempt := 0; attempt < retries; attempt++ {
+retryLoop:
+	for attempt := 0; attempt < retries; attempt++ {
+		time.Sleep(retryInterval)
+		response, err := client.Nrdb.Query(envutil.GetNrAccountId(), query)
+		if err != nil {
+			t.Log(err)
+			continue retryLoop
+		}
+		if len(response.Results) != len(m.Assertions) {
+			t.Fatalf("Query results (%+v) and number of assertions (%+v) do not match", response.Results, m.Assertions)
+		}
+		for i, assertion := range m.Assertions {
 			actualContainer := response.Results[i]
 			actualValue := actualContainer[assertion.AggregationFunction+"."+m.Query.Metric.Name]
 			valueType := reflect.TypeOf(actualValue)
 			if valueType == nil || valueType.Kind() != reflect.Float64 {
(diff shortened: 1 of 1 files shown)


test/e2e/util/assert/nr_metric_assertion.go
```diff
@@ -63,26 +63,28 @@ type NrAssertion struct {
...
+			t.Fatalf("Query results (%+v) and number of assertions (%+v) do not match", response.Results, m.Assertions)
+		}
+		for i, assertion := range m.Assertions {
 			actualContainer := response.Results[i]
 			actualValue := actualContainer[assertion.AggregationFunction+"."+m.Query.Metric.Name]
 			valueType := reflect.TypeOf(actualValue)
 			if valueType == nil || valueType.Kind() != reflect.Float64 {
 				t.Logf("Attempt %d: Expected float64 for assertion %+v, but received %+v in response %+v. Retrying...", attempt, actualContainer, valueType, response.Results)
-				continue
+				continue retryLoop
```
mailo-nr, line 83: Am I reading this correctly that if this is the last iteration we would report a false positive?

test/e2e/util/assert/nr_metric_assertion.go
```diff
@@ -63,26 +63,28 @@ type NrAssertion struct {
...
-		t.Fatal(err)
-	}
-	if len(response.Results) != len(m.Assertions) {
-		t.Fatalf("Query results (%+v) and number of assertions (%+v) do not match", response.Results, m.Assertions)
-	}
-	for i, assertion := range m.Assertions {
-		for attempt := 0; attempt < retries; attempt++ {
+retryLoop:
+	for attempt := 0; attempt < retries; attempt++ {
+		time.Sleep(retryInterval)
```
mailo-nr, line 68: Could add a check to see if attempts is > 0 as to not sleep on the first try.

test/e2e/util/assert/nr_metric_assertion.go
kb-newrelic: You're correct, added some more robust logic by keeping track with a successful assertion counter.

test/e2e/util/assert/nr_metric_assertion.go
```diff
@@ -63,26 +63,37 @@ type NrAssertion struct {
...
-		for attempt := 0; attempt < retries; attempt++ {
+retryLoop:
+	for attempt := 0; attempt < retries; attempt++ {
+		if attempt > 0 {
+			time.Sleep(retryInterval)
+		}
+		response, err := client.Nrdb.Query(envutil.GetNrAccountId(), query)
+		if err != nil {
+			t.Log(err)
+			continue retryLoop
```
mailo-nr, line 74: Same issue here yeah? An error on the last iteration, the loop finishes and no error thrown.

mailo-nr (dismissed):
LGTM

test/e2e/util/assert/nr_metric_assertion.go
```diff
@@ -62,29 +64,48 @@ type NrAssertion struct {
...
 func (m *NrMetricAssertion) ExecuteWithRetries(t testing.TB, client *newrelic.NewRelic, retries int, retryInterval time.Duration) {
+	for attempt := 0; attempt < retries; attempt++ {
+		if attempt > 0 {
+			time.Sleep(retryInterval)
+		}
+		err := m.execute(t, client)
+		if err == nil {
+			break
+		} else {
+			t.Logf("Assertion attempt %d failed with: %+v", attempt, err)
```
mailo-nr, line 75: Do we need to add something like:
```
if (attempt + 1 >= retries) {
    t.Fatalf...
}
```

mailo-nr (approved):
LGTM

PR #238, opened 2025-02-21
fix: provide namespace for clusterrole in nightly

kb-newrelic:
### Summary
- `namespace` value is missing from nightly helm chart

### DIFF
--- a/test/charts/mocked_backend/values.yaml
+++ b/test/charts/mocked_backend/values.yaml
@@ -3,5 +3,3 @@ image:
   tag: latest
   # Avoid accidentally pulling remote images in CI
   pullPolicy: Never
-
-namespace: default-mocked-backend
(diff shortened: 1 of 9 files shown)


test/charts/mocked_backend/values.yaml
```diff
@@ -4,4 +4,4 @@ image:
   # Avoid accidentally pulling remote images in CI
   pullPolicy: Never
 
-namespace: default-mocked-backend
+namespace: mocked-backend-default-namespace
```
mailo-nr, line 7: Helm should have the builtin `{{ .Release.Name }}` to use in templates that references the namespace set in the helm release

kb-newrelic: Good point, that's much cleaner! [Looks like](https://helm.sh/docs/chart_template_guide/builtin_objects/) it is `.Release.Namespace`

kb-newrelic: Removed the code that was necessary to pass it around in tests as well.

mailo-nr (approved):
LGTM

PR #242, opened 2025-02-21
ci: limit concurrency of nightly tf

kb-newrelic:
### Summary
- Having nightly apply the tf for each distro concurrently causes deadlocks in dynamoDB and state corruption when one tf fails and kills the other one.
- Following [github docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-using-concurrency-and-the-default-behavior) using concurrency groups to limit concurrency to 1


mailo-nr (approved):
LGTM

PR #243, opened 2025-02-21
ci: add label to goreleaser publish step

kb-newrelic:
### Summary
- goreleaser nightly publish step was missing an `id` to extract the artifacts


mailo-nr (approved):
LGTM

PR #249, opened 2025-02-25
test: add test support for k8s metrics

kb-newrelic:
### Summary
- Extend test architecture to allow testing different set of metrics for different distributions
   - new concept of a `test case spec` which is a set of metrics and conditions on those metrics to be tested
      - existing tests became the `host` test case spec and one example test was added for a `k8s` test case spec (latter name might still change when we add more test cases and decide how to properly group them)
   - `test spec` was extended to define list of `test case specs` to be by type of test (slow and nightly can have different set of metrics to be tested)
- Update helm charts to support k8s distro in 'deployment mode' for tests and nightly
   - deployment of deployment collector is only enabled for the k8s distro
   - `clusterName` was added as input to helm chart to populate `k8s.cluster.name`

### Misc

- Namespace test clusters by run_id and attempt to ensure that reruns of a workflow do not unintentionally reuse dangling resources

### Notes
- I am open to use a different name for `test spec` vs `test case spec`, didn't want to dwell on it too long for now as it is easily changeable


test/charts/mocked_backend/templates/deployment-collector.yaml
```diff
@@ -0,0 +1,40 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: &app nrdot-collector-deployment
+  labels:
+    app: *app
+spec:
+  replicas: {{ .Values.image.repository | hasSuffix "k8s" | ternary 1 0 }}
```
mailo-nr, line 9: We could probably conditionally render the whole deployment if it isn't needed

mailo-nr (approved):
Added a couple comments but nothing worth holding this up.

test/charts/mocked_backend/templates/deployment-collector.yaml
kb-newrelic: Good point. I'll incorporate that into the follow-up PR that adds the remaining metrics were testing for

PR #251, opened 2025-02-25
test: adds deps for entity synthesis

kb-newrelic:
### Summary
- Deploy kube-state-metrics dependency to ensure that metrics with metadata necessary for entity synthesis can be scraped, inspired by [nr-k8s-otel-collector](https://github.com/newrelic/helm-charts/blob/master/charts/nr-k8s-otel-collector/Chart.yaml#L26-L29)
- Update permissions to those used by [nr-k8s-otel-collector](https://github.com/newrelic/helm-charts/blob/master/charts/nr-k8s-otel-collector/templates/clusterrole.yaml) as we were missing one required to synthesize the cluster entity (and potentially more). My initial somewhat misled goal was to have a minimal set of permissions based on the otel docs but the k8s team did exactly that work, so no need to redo that.

### Misc
- Conditionally render all resources that are only necessary for the `k8s` distro
- Remove unused dependencies from the `mock_backend` chart. I initially included them to keep both charts in sync but the more we add, the more it just becomes a chore without providing any real value. Once we actually evolve the local test to test for specific metrics, we can add them back in.


mailo-nr (approved):
LGTM

PR #252, opened 2025-02-26
ci: use shorter cluster names compatible with kind

kb-newrelic:
### Summary
- [Nightly failed](https://github.com/newrelic/nrdot-collector-releases/actions/runs/13546938296/job/37860805144#step:14:31) due to cluster name being too long when I added the run-id + attempt, so removing unnecessary prefixes to hopefully make it work. Most likely limit we're hitting [is 63](https://stackoverflow.com/a/28918017) (couldn't find anything official from kind/docker) as the [docker run](https://github.com/newrelic/nrdot-collector-releases/actions/runs/13546938296/job/37860805144#step:14:38) with hostname `ci-e2etest-nightly-nrdot-collector-host-13546938296-1-control-plane` failed (67 chars).

### DIFF
--- a/.github/workflows/ci-base.yaml
+++ b/.github/workflows/ci-base.yaml
@@ -13,7 +13,7 @@ on:
       test_cluster_name:
         required: false
         type: string
-        default: 'ci-e2etest-${{ github.run_id }}-${{ github.run_attempt }}'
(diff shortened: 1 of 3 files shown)


mailo-nr (approved):
LGTM

PR #255, opened 2025-02-27
ci: initialize map to avoid NPE

kb-newrelic:
### Summary
- fixes [nightly error](https://github.com/newrelic/nrdot-collector-releases/actions/runs/13569448551/job/37932129773#step:4:27)


mailo-nr (approved):
LGTM

PR #257, opened 2025-02-27
test: remove request spacing behavior in tests

kb-newrelic:
### Summary
- With query retry in place, we no longer need to artificially space out requests and can instead just run into the rate limit and retry to keep it simple. It also speeds up nightly tests a lot (1min vs 5s locally)
- This should help identify the root cause of [nightly timing out](https://github.com/newrelic/nrdot-collector-releases/actions/runs/13572948183/job/37943719544#step:4:32)


mailo-nr (approved):
LGTM

PR #258, opened 2025-02-28
ci: fix host.name for host distro in nr_backend

kb-newrelic:
### Summary
- Fixes that the helm conditionals for the k8s cluster unintentionally also messed up the host name for the host distro.
- Also add conditionals for k8s-only env vars for the mock_backend chart
image


mailo-nr (approved):
LGTM

PR #259, opened 2025-02-28
ci: sign bot user commits

mailo-nr:
Ensure release tags generated in ci are signed.

### DIFF
--- a/.github/workflows/release-tag.yaml
+++ b/.github/workflows/release-tag.yaml
@@ -16,6 +16,14 @@ jobs:
           fetch-depth: 0 # required for tag metadata
           token: ${{ secrets.RELEASE_TOKEN }} # Uses a PAT to ensure subsequent workflows get triggered
 
+      - name: Import GPG key
+        uses: crazy-max/ghaction-import-gpg@v6
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-tag.yaml
```diff
@@ -16,6 +16,14 @@ jobs:
           fetch-depth: 0 # required for tag metadata
           token: ${{ secrets.RELEASE_TOKEN }} # Uses a PAT to ensure subsequent workflows get triggered
 
+      - name: Import GPG key
+        uses: crazy-max/ghaction-import-gpg@v6
+        with:
+          gpg_private_key: ${{ secrets.OTELCOMM_BOT_GPG_PRIVATE_KEY_BASE64 }}
+          passphrase: ${{ secrets.OTELCOMM_BOT_GPG_PASSPHRASE }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
```
kb-newrelic, line 25: there is also the option `git_tag_gpgsign`, not sure if commit implies tag or if it is a separate option ?

kb-newrelic: Oh I see, the `git tag -s` option makes it explicit, so this probably works :)

kb-newrelic (approved):

PR #262, opened 2025-03-01
ci: set cluster name for nightly collectors

kb-newrelic:
### Summary
- Set clustername for nightly helm chart to avoid default cluster name 
image


mailo-nr (approved):
LGTM

PR #270, opened 2025-03-10
ci: reduce cloudwatch retention for eks to 1d

kb-newrelic:
### Summary
- Reduce cost by limiting cloudwatch retention for the EKS cluster. We don't really use the logs, so reduced it to 1d in case we want to debug something live.


mailo-nr (approved):
💸

PR #272, opened 2025-03-12
docs: add troubleshooting guide

kb-newrelic:
### Summary
- Add general and distro-specific troubleshooting docs

### DIFF
--- a/distributions/TROUBLESHOOTING.md
+++ b/distributions/TROUBLESHOOTING.md
@@ -1,0 +1,114 @@
+# NRDOT Collector Troubleshooting
+Please make sure to consult the [README](./README.md) first to ensure that you have the necessary prerequisites in place.
+The collector is extremely powerful but therefore also easily misconfigured. As we cannot cover all the customization possibilities, we recommend consulting the [OTel Collector troubleshooting documentation](https://opentelemetry.io/docs/collector/troubleshooting/) if this guide does not help you resolve your issue. At its core, NRDOT is a [custom distribution](https://opentelemetry.io/docs/collector/custom-collector/), so many of the same steps will apply.
+
+This guide covers general collector troubleshooting tools and common issues for. Distribution-specific troubleshooting is available here:
+- [nrdot-collector-host](./nrdot-collector-host/TROUBLESHOOTING.md)
+- [nrdot-collector-k8s](./nrdot-collector-k8s/TROUBLESHOOTING.md)
+
+However, note that those guides assume you are familiar with the tools mentioned in this guide.
+
+## Helpful Tools
+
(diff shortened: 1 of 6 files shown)


distributions/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,114 @@
+# NRDOT Collector Troubleshooting
+Please make sure to consult the [README](./README.md) first to ensure that you have the necessary prerequisites in place.
+The collector is extremely powerful but therefore also easily misconfigured. As we cannot cover all the customization possibilities, we recommend consulting the [OTel Collector troubleshooting documentation](https://opentelemetry.io/docs/collector/troubleshooting/) if this guide does not help you resolve your issue. At its core, NRDOT is a [custom distribution](https://opentelemetry.io/docs/collector/custom-collector/), so many of the same steps will apply.
+
+This guide covers general collector troubleshooting tools and common issues for. Distribution-specific troubleshooting is available here:
```
mailo-nr, line 5: typo after `for` looks like

distributions/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,114 @@
...
+Error: failed to get config: cannot unmarshal the configuration: decoding failed due to the following error(s):
+error decoding 'processors': error reading configuration for "memory_limiter": decoding failed due to the following error(s):
+'limit_mib' expected type 'uint32', got unconvertible type 'string', value: '100m'
+```
+
+### Telemetry not reaching New Relic
+If the UI does not light up as you expect it to and there are no [NrIntegrationError](https://docs.newrelic.com/docs/data-apis/ingest-apis/metric-api/troubleshoot-nrintegrationerror-events/), you can run some basic NRQL queries to check whether telemetry is reaching New Relic at all. If you do see some but not all the data you expect, there is either an issue with a specific pipeline or you might be running into [cardinality limits](https://docs.newrelic.com/docs/data-apis/ingest-apis/metric-api/NRQL-high-cardinality-metrics/).
+```
+# Metrics
+FROM Metric SELECT * WHERE newrelic.source='api.metrics.otlp' WHERE otel.library.name like 'github.com/open-telemetry/opentelemetry-collector-contrib/receiver%' SINCE 1 hour ago
```
mailo-nr, line 96: Should we add a condition on the user agent?  Wondering if customers might have data from other sources that would give them false-positives.

kb-newrelic: Would love to but as far as I can tell, the user agent is not accessible by customers. We only have internal access to it as the ingest service records it on its own telemetry.

mailo-nr (approved):
LGTM

PR #274, opened 2025-03-13
fix: add missing configs to archive

kb-newrelic:
### Summary
- Add default collector configs to archives that are currently still missing


mailo-nr (approved):
LGTM

PR #280, opened 2025-03-17
chore: remove unnecessary init container

kb-newrelic:
### Summary
- Remove unnecessary `initContainers` for nr_backend chart as it does not contain the validation backend - was accidentally copied from mock-backend


mailo-nr (approved):
LGTM

PR #292, opened 2025-03-25
docs: add missing documentation

kb-newrelic:
### Summary
- Update README and troubleshooting docs.

### Misc
- Added invisible anchor for heading with longer name not unlikely to change later on to avoid broken links. Github seems to prefix ids with `user-content`, so this is a bit hacky but IMO still less likely to break than referencing by plain name for headers that are not standardized like 'Configuration' or 'Installation'.


mailo-nr (approved):
LGTM

PR #297, opened 2025-04-10
chore: add new workflow to notify if nrdot artifacts match AC expectations

alvarocabanas:
Workflow triggered on commit or PR to main that checks the integrity of the artifacts used by Agent Control when embedding the NRDOT on host binary and config.

Checks:
- Ensure config.yaml file exists in current path
- Check for config.yaml is not modified in last commit.
- Check binary name in goreleaser file matches expected.
- Check package name in goreleaser file matches expected.

If any of this checks doesn't pass a message is sent to AC and nrdot slack channels to be reviewed. But the workflow doesn't fail.

### DIFF
--- a/.github/workflows/check-artifacts-ac.yaml
+++ b/.github/workflows/check-artifacts-ac.yaml
@@ -1,0 +1,105 @@
+name: 🔄 Check artifacts to match AC expectations
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  check-artifacts-ac:
+    name: Check artifacts to match AC expectations
+    runs-on: ubuntu-latest
+    outputs:
+      checks: ${{ steps.aggregated_msg.outputs.failure_message }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
+
+      - name: Ensure config.yaml file exists
+        id: check_file_existence
+        run: |
+          if [ ! -f distributions/nrdot-collector-host/config.yaml ]; then
+            echo "file_exists=false" >> $GITHUB_ENV
+            echo "file_message=The file distributions/nrdot-collector-host/config.yaml does not exist." >> $GITHUB_ENV
+            exit 0
+          fi
+          echo "file_exists=true" >> $GITHUB_ENV
+
+      - name: Check for config.yaml modification
+        id: check_config
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/check-artifacts-ac.yaml
```diff
@@ -0,0 +1,105 @@
...
+          fi
+          if [ "$binary_correct" == "false" ]; then
+            failure_message="$failure_message\n$binary_message"
+          fi
+          if [ "$package_correct" == "false" ]; then
+            failure_message="$failure_message\n$package_message"
+          fi
+          echo "failure_message=$failure_message" >> $GITHUB_OUTPUT
+
+  notify-conflicts:
```
gsanchezgavier, line 82: why not making the workflow fail instead of notifying ?
Is it desirable that we brake this contract?

kb-newrelic: I agree failing a PR for violating the basic requirements (path, binary/package name) is reasonable. As we already have a sync/check for the k8s distro, I created a [PR](https://github.com/newrelic/nrdot-collector-releases/pull/298/files#diff-dd30ed49521c1f000579f85c757adb6e69e44b12daef43b1e3c16c9faee54d3c) to follow the same pattern but it was super helpful to have yours as a template, so I appreciate the work you put in!

As you have already pointed out, we can't have the same strictness for the config as it would create a pretty limiting dependency.

I was hoping we could codify the dependency on the config a bit more clearly to avoid having to start the analysis from scratch anytime this config changes. Like if we were to change the `memorylimiterprocessor` defaults, I assume it would be unnecessary to alert either one of our teams but I need to understand the exact dependency better to codify it properly.
Things that come to mind would be:
- any change in env vars names (removal, newly added, renames)
- host entity synthesis (out of scope for this - it's in our backlog to validate that in our E2E tests though)
- ...?

If your team wants to be notified of any change, we can for sure do that but for our team it would be a non-actionable alert, so I would like to avoid that.

kb-newrelic: We can decide in this thread what to do and I will replace [this TODO](https://github.com/newrelic/nrdot-collector-releases/pull/298/files#r2043103385) with whatever we end up with.

kb-newrelic:
Closing this in favor of #298

.github/workflows/check-artifacts-ac.yaml
kb-newrelic: Added env var assertions: https://github.com/newrelic/nrdot-collector-releases/pull/298/commits/4e94d7f5e781f6a35ea257e28aff478e5c6f6057

PR #298, opened 2025-04-14
ci: add agent control config assertion workflow

kb-newrelic:
### Summary
- PR in similar vein as #297:
> Checks:
> Ensure config.yaml file exists in current path
> Check binary name in goreleaser file matches expected.
> Check package name in goreleaser file matches expected.
- [x] We need to figure out a version of 'Check for config.yaml is not modified in last commit.' that allows us to still  evolve the host distro if necessary, see [this discussion](https://github.com/newrelic/nrdot-collector-releases/pull/297#discussion_r2043101358)


.github/workflows/ci-nightly-k8s-config-drift.yaml
```diff
@@ -0,0 +1,23 @@
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v4.2.0
+
+      - name: Sync k8s distro configs from helm chart
+        run: |
+          make -f ./distributions/nrdot-collector-k8s/Makefile check-k8s-configs
```
kb-newrelic, line 23: Renaming messed up the diff - the only change is using `check-k8s-configs` instead of `sync-configs` as the latter doesn't fail on drift.

Makefile.dev
```diff
@@ -1,10 +1,10 @@
 .PHONY: ci
 ci:
-	act -W .github/workflows/ci.yaml
+	act push -W .github/workflows/ci.yaml
```
kb-newrelic, line 3: Took me a while to figure out why just changing the `event.json` didn't affect the variable `${{github.event_name}}`. Figured I make it more obvious that `act -W ...` default to `push`.

event.json
```diff
@@ -1,3 +1,13 @@
 {
-  "act": true
+  "act": true,
+  "pull_request": {
```
kb-newrelic, line 3: Example payload for pull_request event, see [github docs](https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request) and [act docs](https://nektosact.com/usage/index.html#using-event-file-to-provide-complete-event-payload). Doesn't seem to affect anything when running with `act push`, so I just left it in.

distributions/nrdot-collector-host/assert-invariants-agent-control.sh
```diff
@@ -0,0 +1,44 @@
...
+  'NEW_RELIC_LICENSE_KEY'
+)
+for env_var in "${env_vars[@]}"; do
+  echo "Checking for env var ${env_var}"
+  echo "\${env:${env_var}[^}]+}"
+  grep -E '\${env:[^}]+}' "${host_distro_dir}/config.yaml" |
+  grep "${env_var}" ||
+    { echo "expected env var '${env_var}' in config.yaml"; exit 1; }
+done
+# TODO: assert additional invariants within config
```
kb-newrelic, line 27: Add additional assertions if necessary

distributions/nrdot-collector-host/assert-invariants-agent-control.sh
```diff
@@ -0,0 +1,44 @@
...
+
+
+# assert binary name
+goreleaser_yamls=('.goreleaser.yaml' '.goreleaser-nightly.yaml')
+for goreleaser_yaml in "${goreleaser_yamls[@]}"; do
+  echo "Checking ${goreleaser_yaml}"
+  yq -e '.builds[].binary == "nrdot-collector-host"' "${host_distro_dir}/${goreleaser_yaml}" ||
+    { echo "expected binary name 'nrdot-collector-host' in ${goreleaser_yaml}"; exit 1; }
```
alvarocabanas, line 35: Awesome that you fail the pipeline, we are gonna ensure we don't break the contract.

distributions/nrdot-collector-host/assert-invariants-agent-control.sh
alvarocabanas: Does it make sense to add an asertion to ensure we are detecting and setting the host.id resource needed to create the host entity?

```
  resourcedetection:
    detectors: ["system"]
    system:
      hostname_sources: ["os"]
      resource_attributes:
        host.id:
          enabled: true

```

kb-newrelic: done: https://github.com/newrelic/nrdot-collector-releases/pull/298/commits/cd143b53ebc149fa5e0a330ae8297f9656d816b6#diff-dd30ed49521c1f000579f85c757adb6e69e44b12daef43b1e3c16c9faee54d3cR27-R29

mailo-nr (approved):
Looks reasonable to me 👍

PR #303, opened 2025-05-01
Updating otel components to 0.125.0

mailo-nr:


mailo-nr:
The k8s team actually requested that we enable the feature gate early so we should be good for this release, I do think it's a good idea to include the relevant component changes in the changlog.

kb-newrelic:
Created https://new-relic.atlassian.net/browse/NR-410138 to follow-up on the changelog idea

kb-newrelic (approved):
LGTM

PR #307, opened 2025-05-06
ci: ensure conventional commit PR title

kb-newrelic:
### Summary
- Simple workflow to validate PR title to avoid the suggested commit message to not adhere to conventional commits and therefore end up excluded from the auto-generated changelogs in the draft release.
- If it fails, it comments (like [here](https://github.com/newrelic/nrdot-collector-releases/pull/307#issuecomment-2855943630) when I removed the `ci:` prefix from the title) and fails the workflow


mailo-nr (approved):
LGTM

PR #309, opened 2025-05-13
feat: Add cli docker snapshot release

mailo-nr:
Adds a workflow to build and publish a snapshot of the nrdot-collector-builder to ghcr.io
Tests run but will not fail the build until stabilized

### DIFF
--- a/cmd/nrdot-collector-builder/.goreleaser.yaml
+++ b/cmd/nrdot-collector-builder/.goreleaser.yaml
@@ -1,0 +1,97 @@
+version: 2
+builds:
+  - flags:
+      - -trimpath
+    ldflags:
+      - -s
+      - -w
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - windows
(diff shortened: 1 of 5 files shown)


cmd/nrdot-collector-builder/.goreleaser.yaml
```diff
@@ -0,0 +1,97 @@
...
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - windows
+      - darwin
+    goarch:
+      - amd64
+      - arm64
+      - ppc64le
```
kb-newrelic, line 17: is that an arch we use anywhere? I wasn't aware

kb-newrelic (approved):

cmd/nrdot-collector-builder/.goreleaser.yaml
mailo-nr: Not currently, but I figured we might as well start with parity with OCB

PR #315, opened 2025-05-19
ci: update otel components on cron

mailo-nr:
Check if there are new upstream otel component versions and pull a pr if necessary.

### DIFF
--- a/.github/workflows/ci-bump-component-versions.yaml
+++ b/.github/workflows/ci-bump-component-versions.yaml
@@ -1,0 +1,77 @@
+name: 🔄 CI | Bump Component Versions
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Scheduled to run every hour on every day-of-week from Monday through Friday.
+    - cron: '0 * * * 1-5'
+
+jobs:
+    bump:
+        name: Bump Component Versions
+        runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-bump-component-versions.yaml
```diff
@@ -0,0 +1,77 @@
...
+            git add --all
+            git commit -m "feat: Bump otel component versions from ${{ env.current_beta_core }} to ${{ env.next_beta_core }}"
+            git push origin ${{ env.branch }}
+            
+        - name: Issue PR
+          if: ${{ !env.ACT && steps.bump_component_versions.outputs.has_changes == 'true' }}
+          env:
+            GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
+          run: |
+            gh pr create --title "Bump OTEL beta core to ${{ env.next_beta_core }}" \
```
kb-newrelic, line 74: Should we check beforehand if that PR already exists? Otherwise I would assume this to either create duplicates or fail?

mailo-nr: Yeah, plan is to iterate on it.  We should probably close out old prs from versions that weren't merged as well

kb-newrelic (approved):

PR #317, opened 2025-05-20
fix: Fix pr generation on version bump

mailo-nr:
Needed to fix a few things around git commits, also improved a few things:

- Handles cases where remote branch already exists
- Closes prs for older prs that are still open
- Updates pr if it already exists

Here is a link to the successful run: https://github.com/newrelic/nrdot-collector-releases/pull/316

### DIFF
--- a/.github/workflows/ci-bump-component-versions.yaml
+++ b/.github/workflows/ci-bump-component-versions.yaml
@@ -6,6 +6,10 @@ on:
     # Scheduled to run every hour on every day-of-week from Monday through Friday.
     - cron: '0 * * * 1-5'
 
+permissions:
+  contents: write
+  pull-requests: write
+
 jobs:
     bump:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-bump-component-versions.yaml
```diff
@@ -15,6 +19,20 @@ jobs:
           uses: actions/checkout@v4
           with:
             fetch-depth: 0 # required for tag metadata
+            token: ${{ secrets.RELEASE_TOKEN }}
```
kb-newrelic, line 22: curious why this is necessary?

mailo-nr: If I set it here I don't need to set it on each step

kb-newrelic (approved):

PR #325, opened 2025-06-24
Kbauer/debundle k8s configs

kb-newrelic:
### Summary
- Preparation for removing default config from k8s distro
   - update docs to be consistent with no default config for k8s distro
   - migrate fast and slow tests to use [nr-k8s-otel-collector](https://github.com/newrelic/helm-charts/tree/master/charts/nr-k8s-otel-collector) chart instead of our custom chart which would still require the default config
   - disable fast tests for k8s config as there is no real meaningful way to run these without a default config

### Notes
- We should consider running slow tests in CI and not only in nightly and possibly remove the fast tests completely. It would increase CI time but also prevent merging broken configs as it happened for the last k8s config sync
-


test/e2e/slow/local_collector_nr_backend_test.go
```diff
@@ -53,7 +54,6 @@ func TestSlow(t *testing.T) {
 		whereClause := testCaseSpec.RenderWhereClause(testEnvironment)
 		t.Logf("test case spec where clause: %s", whereClause)
 
-		counter := 0
```
kb-newrelic, line 56: unused code

test/e2e/nightly/nightly_collector_nr_backend_test.go
```diff
@@ -51,7 +51,6 @@ func TestNightly(t *testing.T) {
 			}
 
 			whereClause := testCaseSpec.RenderWhereClause(testEnvironment)
-			counter := 0
```
kb-newrelic, line 54: unused

mailo-nr (approved):
LGTM

PR #326, opened 2025-06-25
test: use nr-k8s-otel-collector for k8s nightly

kb-newrelic:
### Summary
- Use nr-k8s-otel-collector helm chart for nightly tests of k8s distro
- Peer PR to [this one](https://github.com/newrelic/nrdot-collector-releases/pull/325) which addresses fast and slow tests

PR #327, opened 2025-06-25
feat: remove default config from k8s distro

kb-newrelic:
### Summary (merge after #325 + #326)
- Removes default configs from k8s distro
- Removed all other references to the config
   - documentation
   - nr_backend helm chart (removed all conditional logic around k8s distro as well as #325 and #326 remove that dependency)


mailo-nr (approved):
LGTM

PR #336, opened 2025-06-27
chore(deps): ensure dep chart version lockstep

kb-newrelic:
### Summary
- Renovate doesn't match same regex twice per file (#334) but rather just the last one as far as I can tell, otherwise I would've expected a second PR by now.
- The best I could come up with to address this was to use yaml anchors instead. It also expresses more clearly that those versions are supposed to be the same
- Reverting it back to 0.8.28 to see if renovatebot works correctly

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -13,7 +13,7 @@
       ],
       "matchStringsStrategy": "combination",
       "matchStrings": [
-        ".*newrelic/(?<depName>nr-k8s-otel-collector)\\s+version:\\s+(?<currentValue>[\\d\\.]+)",
+        ".*# renovatebot.*\\s*name:\\s+newrelic\/(?<depName>nr-k8s-otel-collector)\\s+version:\\s+&k8s_chart_version\\s+(?<currentValue>[\\d\\.]+)",
       ],
       "registryUrlTemplate": "https://helm-charts.newrelic.com",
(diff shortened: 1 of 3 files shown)


.github/renovate.json5
```diff
@@ -13,7 +13,7 @@
       ],
       "matchStringsStrategy": "combination",
       "matchStrings": [
-        ".*newrelic/(?<depName>nr-k8s-otel-collector)\\s+version:\\s+(?<currentValue>[\\d\\.]+)",
+        ".*# renovatebot.*\\s*name:\\s+newrelic\/(?<depName>nr-k8s-otel-collector)\\s+version:\\s+&k8s_chart_version\\s+(?<currentValue>[\\d\\.]+)",
```
mailo-nr, line 16: Is there any way we can use something like:
```
".*# renovatebot.*\\s*name:\\s+(?<depName>.*?)\\s+version:\\s+(?:&[\\w_]+\\s+)?(?<currentValue>[\\d\\.]+)",
```
So we can ignore the anchor?

kb-newrelic: I incorporated your suggestion. I also added a workflow that validates that renovate actually matches the file and dependency. This should ensure that we don't accidentally break either part of the 'equation' without noticing that we broke renovate.

PR #338, opened 2025-06-27
chore: remove fast tests

kb-newrelic:
### Summary
- Since #328 we run slow tests for PRs, so no need for fast tests which have a smaller scope / provide less confidence


mailo-nr (approved):
LGTM

PR #345, opened 2025-07-09
test: add internal telemetry tests

kb-newrelic:
### Summary
- Add tests for internal telemetry based on the [newrelic/newrelic-integration-e2e-action](https://github.com/newrelic/newrelic-integration-e2e-action)
- Add example configuration for internal telemetry until we integrate it as default behavior for NRDOT. The example config is used 'as-is' for the e2e tests.
- E2E tests limited to host distro as behavior shouldn't differ across distros and host distro is the easiest one to tailor towards test requirements
   - Various internal metrics require certain input (generated by telemetrygen) or component configuration to be generated in the first place
- Collector is deployed via the [opentelemetry-collector](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-collector) helm chart

### Notes
- Follow-up PR will migrate the normal telemetry tests for both distros to the new action and remove slow tests and thus the necessity for us to maintain our custom test code.


.github/workflows/ci-e2e.yaml
```diff
@@ -0,0 +1,29 @@
...
+      docker_hub_username: ${{ secrets.OTELCOMM_DOCKER_HUB_USERNAME }}
+      docker_hub_password: ${{ secrets.OTELCOMM_DOCKER_HUB_PASSWORD }}
+      gpg_private_key: ${{ secrets.OTELCOMM_GPG_PRIVATE_KEY_BASE64 }}
+      gpg_passphrase: ${{ secrets.OTELCOMM_GPG_PASSPHRASE }}
+      registry: 'newrelic'
+      nr_backend_url: ${{ secrets.NR_STAGING_BACKEND_URL }}
+      nr_ingest_key: ${{ secrets.OTELCOMM_NR_INGEST_KEY }}
+      nr_account_id: ${{ secrets.OTELCOMM_NR_TEST_ACCOUNT_ID }}
+      nr_api_base_url: ${{ secrets.NR_STAGING_API_BASE_URL }}
+      nr_api_key: ${{ secrets.OTELCOMM_NR_API_KEY }}
```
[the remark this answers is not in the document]
kb-newrelic: added it manually, suggestion doesn't go away though

distributions/nrdot-collector-host/test/spec-local.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - cp ../../../examples/internal-telemetry-config.yaml ./relay
+      - kubectl create configmap 'collector-config' --from-file=relay --namespace=nr-${SCENARIO_TAG}
+      - |
+        kubectl create secret generic 'collector-secrets' --namespace=nr-${SCENARIO_TAG} \
+        --from-literal="nrIngestKey=${LICENSE_KEY}" --from-literal="nrBackendUrl=${NR_BACKEND_URL}" \
+        --from-literal="serviceName=${SCENARIO_TAG}-collector" --from-literal="scenarioTag=${SCENARIO_TAG}"
+      - "helm repo add open-telemetry https://open-telemetry.github.io/opentelemetry-helm-charts || echo 'skipping repo add: open-telemetry repo already exists'"
+      - |
+        helm upgrade --install ${SCENARIO_TAG} open-telemetry/opentelemetry-collector \
+        --namespace nr-${SCENARIO_TAG} --version 0.129.0 \
```
mailo-nr, line 19: Can we read this version in somehow so we don't have to worry about drift?

kb-newrelic: I guess we could just align it with the component version and then integrate it in the version bump script. What do you think? The version itself should probably still live in the ci-base.yaml though.

Or did you have something else in mind?

kb-newrelic: Found a decent solution I think. The [otel chart](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/Chart.yaml#L15) has two versions, one for the chart itself and one for the intended collector version (appVersion), so I ended up determining the matching chart version based on the otlpreceiver version in the manifest and as a fallback (in case the collector is quicker bumping versions (and us adopting it) than the helm chart) the latest. 
This should limit breaking changes to version bump PRs which is a good place to address them I think.

mailo-nr (approved):
LGTM, i imagine we can do some re-organizing later if need be.

PR #353, opened 2025-07-23
test: migrate slow tests to e2e action

kb-newrelic:
### Summary
- Replaces slow tests with e2e tests using the e2e-action, similar to how internal telemetry tests were written #345 
   - host distro
      - split into a basic test for existence of metrics and then tailored nrql queries to check for existence of attributes. Unfortunately the e2e-action only supports multiple values in one nrql query if you use facets but facets are not guaranteed to be consistently ordered, so that would make debugging failing tests unnecessarily hard. This is an area where we could contribute and then simplify.
      - extracted common logic for determining the otel collector chart version into a bash script. Technically the nrdot-collector-builder does contain this logic as an internal function, so I could also imagine exposing it through that and then getting rid of the bash script but wanted to avoid bloating this PR even more.
   - k8s distro: ended up deciding to just validate that each deployed collector emits at least one metric. This should validate that an update didn't break something fundamental and any more granular test requires a more intricate setup which makes sense for the nr-k8s-otel-collector chart to own I think.
- Move remaining nightly test spec into `test` folder
- Renovatebot was adjusted to keep helm chart referenced in e2e-action up-to-date


.github/renovate.json5
```diff
@@ -9,11 +9,24 @@
     {
       "customType": "regex",
       "fileMatch": [
-        "(^|/)test-spec.yaml$"
+        "(^|/)spec-nightly.yaml$"
```
mailo-nr, line 12: do we need 2 managers or would 1 work with multiple matches work?

mailo-nr (dismissed):
LGTM

mailo-nr (approved):
LGTM

PR #358, opened 2025-07-30
feat: create fips versions of distributions

emiliaFer:
When you run `make build-fips`, the outcome should be a `_build` and `_build-fips` directory under each distribution.
These directories should match, with the exception that `_build-fips` contains the file `fips.go`.
Then, if you run `goreleaser build --snapshot --clean` and check any of these new directories, you should see new binaries created for each.

### DIFF
--- a/distributions/nrdot-collector-host/.goreleaser.yaml
+++ b/distributions/nrdot-collector-host/.goreleaser.yaml
@@ -5,36 +5,6 @@ release:
   use_existing_draft: true
   disable: "false"
 builds:
-  - id: nrdot-collector-host-fips
-    goos:
-      - linux
-    goarch:
-      - amd64
-      - arm64
-    dir: _build-fips
-    binary: nrdot-collector-host-fips
-    ldflags:
-      - -w
-      - -linkmode external
-      - -extldflags '-static'
-    tags:
-      - netgo
-    flags:
-      - -trimpath
-    env:
-      - CGO_ENABLED=1
-      - GOEXPERIMENT=boringcrypto
-    overrides:
-      - goos: linux
-        goarch: amd64
-        env:
-          - CC=x86_64-linux-gnu-gcc
-          - CXX=x86_64-linux-gnu-g++
-      - goos: linux
-        goarch: arm64
-        env:
-          - CC=aarch64-linux-gnu-gcc
-          - CXX=aarch64-linux-gnu-g++
   - id: nrdot-collector-host
     goos:
       - linux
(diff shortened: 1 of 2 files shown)


distributions/nrdot-collector-host/manifest-fips.yaml
```diff
@@ -0,0 +1,37 @@
+dist:
```
mailo-nr, line 1: Rather than duplicate this I might suggest we copy/modify the manifest to a tmp file before the call to ocb, something like:
```
yq eval '.dist.name += "-fips"' manifest.yaml > manifest-fips.yaml
```

distributions/nrdot-collector-host/.goreleaser.yaml
```diff
@@ -5,6 +5,25 @@ release:
   use_existing_draft: true
   disable: "false"
 builds:
+  - id: nrdot-collector-host-fips
```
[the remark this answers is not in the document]
kb-newrelic: To follow up on this: I suggested adding this target but seeing it now it's probably harder to do so instead of just having a separate file like Mailo mentioned for nightly (I think the limitations of the non-pro versions might also get in the way if not make it impossible).

Makefile
```diff
@@ -22,6 +22,9 @@ ci: check build version-check licenses-check
 check: ensure-goreleaser-up-to-date
 
 build: go ocb
+	@./scripts/build.sh -d "${DISTRIBUTIONS}" -b ${OTELCOL_BUILDER} -s false
```
kb-newrelic, line 25: While this preserves the original state, I think we should be able to get rid of compilation even in this case, i.e. remove `-s false`

cmd/goreleaser/internal/configure.go
```diff
@@ -43,22 +47,27 @@ var (
 	Architectures = []string{"amd64", "arm64"}
 	SkipBinaries  = map[string]bool{
 		K8sDistro: true,
+		K8sDistroFips: true,
 	}
 	NfpmDefaultConfig = map[string]string{
 		HostDistro: "config.yaml",
+		HostDistroFips: "config.yaml",
 		CoreDistro: "config.yaml",
+		CoreDistroFips: "config.yaml",
```
mailo-nr, line 56: If we're using a fips flag we shouldn't need to specific separate distros, fips can just be a suffix

cmd/goreleaser/internal/configure.go
```diff
@@ -267,29 +313,36 @@ func Package(dist string) config.NFPM {
 	}
 }
 
-func DockerImages(dist string, nightly bool) []config.Docker {
+func DockerImages(dist string, nightly bool, fips bool) []config.Docker {
 	var r []config.Docker
+
+	if fips {
+		dist = fmt.Sprint(dist, "-fips")
```
mailo-nr, line 320: I might set this further down the tree so we don't have to worry about bugs related to checks against the distro since we don't need to rely on the distro here and we pass the fips flag to the docker image calls

scripts/build-fips.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+REPO_DIR="$( cd "$(dirname "$( dirname "${BASH_SOURCE[0]}" )")" &> /dev/null && pwd )"
```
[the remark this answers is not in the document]
mailo-nr: did you see my comment related to the yq command?

PR #359, opened 2025-08-04
feat: Introducing nrdot core distribution

mailo-nr:
This PR introduces a new comprehensive `nrdot-collector` distribution that consolidates functionality across host, gateway, and general deployment scenarios. This addresses the current gap in gateway functionality while providing a unified collector binary that can be configured for any deployment pattern.

New Distribution: nrdot-collector
A comprehensive core distribution that includes all universally useful components, enabling configuration-driven differentiation rather than binary restrictions.

Component Analysis vs Host Distribution

```
# Components ADDED to create comprehensive distribution:

receivers:
  + filelogreceiver             # Local file log collection
  + prometheusreceiver          # Application metrics scraping

processors:
  + cumulativetodeltaprocessor  # Metric conversion
  + groupbyattrsprocessor       # Data aggregation  
  + metricstransformprocessor   # Legacy transformations
  + resourceprocessor           # Resource standardization
  + tailsamplingprocessor       # Intelligent sampling

exporters:
  + debugexporter               # Troubleshooting
  + otlphttpexporter           # HTTP variant export
  + loadbalancingexporter      # Backend failover/load balancing

connectors:
  + routingconnector           # Multi-pipeline routing
```

Test specs will be added in a subsequent pr as well as documentation.


distributions/nrdot-collector/config.yaml
```diff
@@ -0,0 +1,37 @@
...
+    endpoint: ${env:OTEL_EXPORTER_OTLP_ENDPOINT:-https://otlp.nr-data.net}
+    headers:
+      api-key: ${env:NEW_RELIC_LICENSE_KEY}
+
+extensions:
+  health_check:
+
+service:
+  extensions: [health_check]
+  pipelines:
```
kb-newrelic, line 25: Note (no change required): I assume this is the beginning of a 'gateway' config. We'll need to keep in mind that if we want the core distro to act as a drop-in replacement for the host distro, the default config would need to be the host config or we make it clear that it's not drop-in and require an additional step of `--config` arguments.

kb-newrelic (approved):

PR #360, opened 2025-08-04
chore: test ci optimizations

mailo-nr:


.github/workflows/ci-base.yaml
```diff
@@ -55,9 +55,34 @@ jobs:
...
         run: go mod tidy
 
-      - name: Verify build
+      - name: Generate cache keys
+        id: cache-keys
+        run: |
+          # Base hash for source generation (inputs that affect generation)
+          SOURCE_HASH="${{ hashFiles('distributions/${{ inputs.distribution }}/manifest.yaml', 'distributions/${{ inputs.distribution }}/_build/go.mod', 'distributions/${{ inputs.distribution }}/_build/go.sum', 'Makefile') }}"
+          # Extended hash for Docker build (includes generated sources)
+          DOCKER_HASH="${{ hashFiles('distributions/${{ inputs.distribution }}/Dockerfile', 'distributions/${{ inputs.distribution }}/manifest.yaml', 'distributions/${{ inputs.distribution }}/_build/go.mod', 'distributions/${{ inputs.distribution }}/_build/go.sum') }}"
```
kb-newrelic, line 64: Embedded configs probably need to be part of the docker hash, right?

.github/workflows/ci-base.yaml
```diff
@@ -92,14 +104,36 @@ jobs:
...
+        if: ${{ !inputs.nightly }}
+        uses: actions/cache@v4
+        with:
+          path: |
+            distributions/${{ inputs.distribution }}/dist
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ steps.build-cache-keys.outputs.binary_key }}
+          restore-keys: |
+            goreleaser-build-${{ inputs.distribution }}-
```
kb-newrelic, line 132: The `binary_key` doesn't have the `goreleaser-build-` prefix, so will this ever hit anything?

.github/workflows/ci-base.yaml
```diff
@@ -111,10 +145,29 @@ jobs:
...
         run: |
-          VERSION=$(echo '${{ steps.goreleaser.outputs.metadata }}' | jq -r '.version')
+          if [ "${{ steps.cache-goreleaser.outputs.cache-hit }}" = "true" ]; then            
+            # Try to find version from dist folder structure or metadata file
+            if [ -f "distributions/${{ inputs.distribution }}/dist/metadata.json" ]; then
+              VERSION=$(jq -r '.version' distributions/${{ inputs.distribution }}/dist/metadata.json)
+            else
+              # Fallback version
+              VERSION="cached-$(date +%Y%m%d-%H%M)"
+              echo "⚠️ Warning: No metadata found, using fallback version: $VERSION"
```
kb-newrelic, line 162: Have we encountered this? Is this to have a fallback when goreleaser changes where it writes its metadata or is there a different scenario I am not thinking of? I am wondering if we should just fail the build instead of semi-silently continue?

kb-newrelic (approved):

.github/workflows/ci-base.yaml
mailo-nr: good catch

PR #394, opened 2025-09-17
chore: exclude binaries from fips distributions

agarvin-nr:
Summary:

- Excludes NFPMs and Blobs from fips distributions


emiliaFer (dismissed):
> Do we even need archives and nfpms? If we can avoid creating those, I would, just to 

> keep things simple. My understanding was the MVP should just be a docker image.

> This would mean we can delete .conf and .service files

Could you remove the fips versions of the `.conf` and `.service` files?

kb-newrelic (approved):
LGTM

PR #396, opened 2025-09-19
Agarvin/ci opt

agarvin-nr:
**Summary**

- Add caching to speed up workflows CI

### DIFF
--- a/.github/workflows/ci-base.yaml
+++ b/.github/workflows/ci-base.yaml
@@ -62,8 +62,10 @@ jobs:
         id: cache-sources
         uses: actions/cache@v4
         with:
-          path: distributions/${{ inputs.distribution }}/_build
-          key: "sources-${{ hashFiles('distributions/${{ inputs.distribution }}/manifest.yaml', 'Makefile') }}"
+          path: |
+            distributions/${{ inputs.distribution }}/_build
+            distributions/${{ inputs.distribution }}/_build-fips
+          key: "sources-${{ hashFiles('distributions/${{ inputs.distribution }}/manifest.yaml', 'distributions/${{ inputs.distribution }}/manifest-fips.yaml', 'Makefile') }}"
 
       - name: Generate sources
         if: steps.cache-sources.outputs.cache-hit != 'true'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-base.yaml
```diff
@@ -112,11 +144,44 @@ jobs:
...
+              format('distributions/{0}/.goreleaser-fips.yaml', inputs.distribution),
+              format('distributions/{0}/_build-fips/*', inputs.distribution)
+            ) }}"
+          else
+            BINARY_HASH="${{ hashFiles(
+              format('distributions/{0}/.goreleaser*.yaml', inputs.distribution),
+              format('distributions/{0}/_build/*', inputs.distribution)
+            ) }}"
+          fi
+          echo "binary_key=goreleaser-build-${{ inputs.distribution }}-${{ github.ref_name }}-${BINARY_HASH}" >> $GITHUB_OUTPUT
```
kb-newrelic, line 165: If we include the ref_name, we'll limit ourselves to caching within the scope of a branch, so basically just reruns of the same PR. One big use case for caching is changes to docs, testing etc., would be cool if we could accommodate those. The hashing should ensure that we rebuild if anything relevant changes.

I know this was copied over from @mailo-nr draft, so curious what his thoughts are as well.

.github/workflows/ci-base.yaml
```diff
@@ -58,9 +58,41 @@ jobs:
...
+
+      - name: Generate sources
+        if: steps.cache-sources.outputs.cache-hit != 'true'
         run: make ci DISTRIBUTIONS=${{ inputs.distribution }}
 
+      - name: Skip source generation (cached)
+        if: steps.cache-sources.outputs.cache-hit == 'true'
+        run: echo "✅ Source generation skipped - no source changes detected"
+
       - name: Login to Docker
```
kb-newrelic, line 96: I wonder if we should add a step that validates that the expected files (independent of whether it was pulled from cache or generated) are present.

.github/workflows/ci-base.yaml
```diff
@@ -112,11 +144,44 @@ jobs:
           elif [ ${{github.event.pull_request.user.login == 'dependabot[bot]' }} ]; then
             echo "goreleaser_args=--snapshot --clean --skip=publish,validate,sign --timeout 2h" >> $GITHUB_ENV
           else
-            echo "goreleaser_args=--snapshot --clean --skip=publish,validate --timeout 2h" >> $GITHUB_ENV
+            echo "goreleaser_args=build --single-target --snapshot --clean" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
kb-newrelic: In addition to the conceptional changes, it might make overall code understanding easier if we have an bool `caching_enabled` (or more granular like `source_caching_enabled`/`binary_caching_enabled`/`image_caching_enabled`) and initialize that with `caching_enabled = !nightly`. That way we clearly indicate the caching related steps and can more easily enable/disable caching for different parts of our CI.

PR #401, opened 2025-09-25
refactor: cleanup CI inconsistencies

kb-newrelic:
### Summary
- Avoid unnecessary source generation in `licenses-check` by checking for existence of `_build` as it was always recompiling sources in CI even thought they were just generated.

- Move around some Makefile targets as we've shifted the compilation completely to goreleaser.
   - remove `build` / `ci` targets and rename dependent targets to more accurately describe what they do: `pre-sourcegen-checks`, `sources`, `post-sourcegen-checks` 
   - only cache the `sources` step, always run `post-sourcegen-checks`
- Makefile refactoring
   - Define Makefile targets with focus on single-distro actions, cross-distro actions have `_all` suffix, e.g. `licenses_all` or `generate-goreleaser_all`
   - mark phony targets as actual phony
   - remove unused target
   - add clean targets
- Fix `ensure-goreleaser-up-to-date` as it isn't checking all goreleaser files due to missing wildcard

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ generate-goreleaser: go
 ensure-goreleaser-up-to-date: generate-goreleaser
 	@{ \
 		git diff -s --exit-code distributions/*/.goreleaser*.yaml || \
-		(echo "📋 Check failed: The goreleaser templates have changed but the .goreleaser.yamls haven't. Run 'make generate-goreleaser' and update your PR." && exit 1) \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-nightly.yaml
```diff
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   schedule:
     # Scheduled to run in the morning (PT) on every day-of-week from Monday through Friday.
-    - cron: '0 15 * * 1-5'
+    - cron: '0 10 * * 1-5'
```
kb-newrelic, line 7: With east coast team members, we want this running a bit earlier.
Should also move the nightly fips workflow up once merged.

PR #403, opened 2025-09-26
chore: add source caching to CI

agarvin-nr:
### Summary

- Adds caching of source files to GitHub workflows
- Fixes caching for local CI by adding a command line argument to .actrc

### Changes to address notes in the [previous PR](https://github.com/newrelic/nrdot-collector-releases/pull/396)
- Adds a global environment variable "caching_enabled" to disable caching (currently disables on nightly)
- Caches sources under a single hash per distribution (instead of needlessly caching for fips and non-fips)
- Includes build.sh in sources cache key hash
- Does not create keys based on ref_name

### Validation

- Both on GitHub and Act: Ran CI workflows twice (one with and one without cache)
- Added step in workflow to explicitly check the existence of source files (both cached and generated)
  - Various local tests deleting build files / directories to see if this would catch problems

### DIFF
--- a/.github/workflows/ci-base.yaml
+++ b/.github/workflows/ci-base.yaml
@@ -72,7 +72,8 @@ jobs:
           SOURCES_HASH="${{ hashFiles(
             format('distributions/{0}/manifest.yaml', inputs.distribution),
             'scripts/build.sh',
-            'Makefile'
+            'Makefile',
+            'fips/fips.go'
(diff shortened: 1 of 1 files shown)


kb-newrelic (approved):
LGTM, we just need to incorporate the fips.go and rebase it

kb-newrelic (approved):

PR #484, opened 2026-01-13
chore: add proprietary license requirements

agarvin-nr:
### Summary
- Adds the following components from the components repo:
  - nrdot-collector
    - receiver/nopreceiver (apache)
  - nrdot-collector-plus
    - receiver/nopreceiver (apache)
    - exporter/nopexporter (proprietary for testing purposes)
- Adds LICENSING file and repository-level LICENSE_ files, similar to the components repo
- Changes third-party notice generation to allow for proprietary licenses
  - Adds override file to recognize component repo
  - Adds third-party notice rules per distro
- Changes goreleaser generation to package licenses with released distributions
- Ports the `nrlicense` tool from the components repo to update license headers

### Validation

- Adds explicit validation of license packaging in archives and packages

### Note

This is a sub-branch to keep a clean distinction between my plus distro changes and the licensing changes. this should probably be merged in with agarvin/plusdistro, and the two merged into main simultaneously


licenses/third_party/overrides.jsonl
agarvin-nr: Ultimately, this file should be the same as the overrides file in the components repo. We may want to consider automatically generating this in components, then downloading it here during the license generation steps.

PR #532, opened 2026-03-16
feat: windows installer

agarvin-nr:
### Summary
- Updates workflows to use GoReleaser Pro (necessary even if skipping msi files, otherwise they can't parse `msi` in the yaml)
- Updates Goreleaser generator to generate MSI files
- Creates MSI installer config file `installer.wxs`, [modeled after OpenTelemetry's config](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/cmd/msi-generator/windows-installer.wxs.tmpl)
- Creates Windows-specific e2e spec to omit Linux-specific host metric queries
- Divides `host-expected-metrics.yaml` into `host-expected-metrics-common.yaml` and `host-expected-metrics-linux.yaml` to support OS-specific test metrics
- Skips MSI package generation in current build pipelines (as not to publish the MSI before it's ready)

### Validation
- Creates workflow `ci-windows` which:
  - Generates an MSI artifact via GoReleaser on an ubuntu runner
  - Passes an artifact into a Windows runner for validation
  - Runs E2E tests
  - Dumps collector logs into console for debugging
  
This workflow was made to run in-parallel with the build pipeline, which is useful for quick iteration. Ultimately, we are going to want to move these scripts to an EC2 - Both for parity with our existing package (debian etc.) validation, and to eventually support additional windows versions.

### DIFF
--- a/.github/workflows/ci-snapshot-cli.yaml
+++ b/.github/workflows/ci-snapshot-cli.yaml
@@ -54,7 +54,7 @@ jobs:
         id: goreleaser
         uses: goreleaser/goreleaser-action@ec59f474b9834571250b370d4735c50f8e2d1e29 # v7
         with:
-          distribution: goreleaser
+          distribution: goreleaser-pro
           version: "~> v2"
           args: --snapshot --clean
           workdir: cmd/nrdot-collector-builder
--- a/.github/workflows/ci-windows.yaml
+++ b/.github/workflows/ci-windows.yaml
@@ -142,6 +142,13 @@ jobs:
 
       - name: Install MSI
         run: |
+          # Set nrdot config environment variables.
+          # Do not need to broadcast WM_SETTINGCHANGE for local env
+          $env:NEW_RELIC_LICENSE_KEY = "${{ secrets.nr_ingest_key }}"
+          $env:OTEL_EXPORTER_OTLP_ENDPOINT = "https://otlp.nr-data.net"
+          $env:OTEL_RESOURCE_ATTRIBUTES = "testKey=$env:SCENARIO_TAG"
+
+          # Install MSI
           $log_path = Join-Path $env:TEMP "msi-install.log"
           $msi_args = @(
               '/i',
@@ -150,24 +157,9 @@ jobs:
               '/l*',
               $log_path
           )
-
-          # Validate install fails when no license key provided
           $process = Start-Process -Wait -PassThru msiexec.exe -ArgumentList $msi_args
-          if ($process.ExitCode -eq 0) {
-            Write-Host "❌ MSI installation allowed to proceed without ingest key"
-            exit 1
-          }
-          if (Test-Path $log_path) {
-            Remove-Item $log_path -Force
-          }
 
-          # Validate successful install on license key provided
-          $msi_args += @(
-            'NEW_RELIC_LICENSE_KEY="${{ secrets.nr_ingest_key }}"',
-            'OTEL_EXPORTER_OTLP_ENDPOINT="https://otlp.nr-data.net"',
-            "OTEL_RESOURCE_ATTRIBUTES=`"testKey=$env:SCENARIO_TAG`""
-          )
-          $process = Start-Process -Wait -PassThru msiexec.exe -ArgumentList $msi_args
+          # Validate install successful
           Write-Host '`nInstallation Log (Last 200 lines):'
           Get-Content $log_path | Select-Object -Last 200
           if ($process.ExitCode -ne 0) {
@@ -180,12 +172,13 @@ jobs:
             exit $process.ExitCode
           }
 
-          # Validate environment variables in registry
-          $env_vars = @('NEW_RELIC_LICENSE_KEY', 'OTEL_EXPORTER_OTLP_ENDPOINT', 'OTEL_RESOURCE_ATTRIBUTES')
-          foreach ($var in $env_vars) {
-            Get-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment' -Name $var | Out-Null
-            Write-Host "✅ $var found in environment registry"
-          }
+          # No registry right now
+          # # Validate environment variables in registry
+          # $env_vars = @('NEW_RELIC_LICENSE_KEY', 'OTEL_EXPORTER_OTLP_ENDPOINT', 'OTEL_RESOURCE_ATTRIBUTES')
+          # foreach ($var in $env_vars) {
+          #   Get-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment' -Name $var | Out-Null
+          #   Write-Host "✅ $var found in environment registry"
+          # }
 
           Write-Host "⏳ Waiting 30 seconds for collector to spool up"
           Start-Sleep -Seconds 30
--- a/.github/workflows/ci-windows.yaml
+++ b/.github/workflows/ci-windows.yaml
@@ -1,0 +1,247 @@
+name: 🧩 CI Windows
+
+on:
+  workflow_call:
+    inputs:
+      distribution:
+        required: true
+        type: string
+    secrets:
+      goreleaser_key:
+        required: true
+      nr_account_id:
+        required: true
+      nr_api_key:
+        required: true
+      nr_ingest_key:
+        required: true
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    outputs:
+      msi_artifact_name: ${{ steps.find-msi.outputs.msi_artifact_name }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # required for tag metadata
+
+      - name: Setup Go
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
+        with:
+          go-version: "~1.26"
+          check-latest: true
+
+      - name: Tidy go.mod files
+        run: go mod tidy
+
+      - name: Ensure goreleaser up to date
+        run: make check
+
+      - name: Generate sources
+        if: steps.cache-sources.outputs.cache-hit != 'true'
+        run: make ci DISTRIBUTIONS=${{ inputs.distribution }}
+
+      - name: Setup wixl # Required to build MSI
+        if: inputs.publish || steps.cache-goreleaser.outputs.cache-hit != 'true'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y wixl
+
+      - name: Generate binary cache key
+        id: generate_binary_key
+        run: |
+          BINARY_HASH="${{ hashFiles(
+            format('distributions/{0}/.goreleaser*.yaml', inputs.distribution),
+            format('distributions/{0}/windows/*', inputs.distribution),
+            format('distributions/{0}/manifest.yaml/*', inputs.distribution)
+          ) }}"
+          echo binary_key=goreleaser-build-${{ inputs.distribution }}-${BINARY_HASH}-windows >> $GITHUB_OUTPUT
+
+      - name: Cache goreleaser build
+        id: cache-goreleaser
+        uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5
+        with:
+          path: |
+            distributions/${{ inputs.distribution }}/dist
+          key: ${{ steps.generate_binary_key.outputs.binary_key }}
+
+      - name: Build binaries & packages with GoReleaser
+        if: steps.cache-goreleaser.outputs.cache-hit != 'true'
+        id: goreleaser
+        uses: goreleaser/goreleaser-action@v6
+        env:
+          GORELEASER_KEY: ${{ secrets.goreleaser_key }}
+          GOOS: windows
+          GOARCH: amd64
+        with:
+          distribution: goreleaser-pro
+          version: "2.11.2"
+          args: "--single-target --snapshot --clean --skip=publish,validate,sign,archive --timeout 2h --config .goreleaser.yaml"
+          workdir: distributions/${{ inputs.distribution }}
+
+      - name: Skip GoReleaser build (cached)
+        if: steps.cache-goreleaser.outputs.cache-hit == 'true'
+        run: echo "✅ GoReleaser build skipped - using cached binaries"
+
+      - name: Find MSI
+        id: find-msi
+        run: |
+          msi_artifact_name=${{ inputs.distribution }}_msi_${{ github.sha }}
+          msi_path="$(find distributions/${{ inputs.distribution }}/dist -name "*.msi" -type f)"
+          if [ ! -f "$msi_path" ]; then
+            echo "❌ Error: .msi not found at $msi_path"
+            exit 1
+          fi
+          echo "msi_path=$msi_path" >> $GITHUB_OUTPUT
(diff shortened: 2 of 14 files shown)


distributions/nrdot-collector/windows/installer.wxs
```diff
@@ -0,0 +1,120 @@
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
+   <Product
+      Name="NRDOT Collector ({{ .Version }})"
+      Id="*"
+      UpgradeCode="BC8657F8-5174-43F3-A319-88D97945291B"
```
kb-newrelic, line 6: I assume we created this uuid?

distributions/nrdot-collector/windows/installer.wxs
```diff
@@ -0,0 +1,120 @@
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
```
kb-newrelic, line 1: Are we intentionally using the v3 schema? I see that the [otel releases](https://github.com/open-telemetry/opentelemetry-collector-releases/pull/572) is using it but that's most likely because it is port from an [earlier implementation](https://github.com/open-telemetry/opentelemetry-collector/blob/feb6cf2403fb2f1a50bfb8f48baa3a7c3ef4452d/internal/buildscripts/packaging/msi/opentelemetry-collector.wxs) that lived in core and was done years ago.

My main argument for migrating would be that this schema url no longer seems to exist and goreleaser pro's [msi docs](https://goreleaser.com/customization/package/msi/) already state that the v3 is not fully supported
> Also note that [msitools](https://wiki.gnome.org/msitools) only supports some parts of the v3 schema.

and while it's a good starting point to use contrib's approach, once we publish the artifact, we're responsible for it.

If the lift is big, let's deal with it in a follow-up.

.github/workflows/ci-snapshot-cli.yaml
```diff
@@ -54,9 +54,9 @@ jobs:
         id: goreleaser
         uses: goreleaser/goreleaser-action@ec59f474b9834571250b370d4735c50f8e2d1e29 # v7
         with:
-          distribution: goreleaser
+          distribution: goreleaser-pro
           version: "~> v2"
-          args: --snapshot --clean
+          args: --snapshot --clean --skip=msi
```
kb-newrelic: shouldn't be necessary as `cmd/nrdot-collector-builder/.goreleaser.yaml` doesn't have an msi section

.github/workflows/release-draft.yaml
```diff
@@ -73,9 +73,9 @@ jobs:
         id: goreleaser
         uses: goreleaser/goreleaser-action@ec59f474b9834571250b370d4735c50f8e2d1e29 # v7
         with:
-          distribution: goreleaser
+          distribution: goreleaser-pro
           version: "2.11.2"
-          args: --clean --timeout 2h
+          args: --clean --timeout 2h --skip=msi
```
kb-newrelic: Top-level goreleaser.yaml shouldn't contain any msi section, so this shoulnd't be necessary

distributions/nrdot-collector/windows/installer.wxs
```diff
@@ -0,0 +1,120 @@
...
+         DowngradeErrorMessage="A later version of NRDOT is already installed. Setup will now exit."/>
+
+      <Feature Id="Feature" Level="1">
+         <ComponentRef Id="ApplicationComponent"/>
+      </Feature>
+
+      <!-- Terminate install if NR license key not provided -->
+      <Condition
+         Message="A New Relic ingest key is required for NRDOT. Please provide NEW_RELIC_LICENSE_KEY during installation.">
+         Installed OR NEW_RELIC_LICENSE_KEY
```
kb-newrelic: Where is this string used? Is this naming an established pattern?

distributions/nrdot-collector/windows/installer.wxs
```diff
@@ -0,0 +1,120 @@
...
+                     Name="{{ .Binary }}"
+                     Start="install"
+                     Stop="both"
+                     Remove="uninstall"
+                     Wait="yes"/>
+
+                  <!-- System environment variables -->
+                  <RegistryKey
+                     Root="HKLM"
+                     Key="SYSTEM\CurrentControlSet\Control\Session Manager\Environment">
```
kb-newrelic: What's the reasoning behind this exact choice? I have no real experience with registry keys, so curious how we got to this value?

distributions/nrdot-collector/windows/installer.wxs
agarvin-nr: Yes! I generated this uuid with `uuidgen`

.github/workflows/ci-snapshot-cli.yaml
agarvin-nr: I missed that this was for nrdot-collector-builder! I'll remove this skip, and the one for top-level

distributions/nrdot-collector/windows/installer.wxs
agarvin-nr: I pieced this together from a few different docs:

- Microsoft's [Environment Variables](https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables) doc describes how there are two different kinds of environment variables (User and System). This doc lists the above key as the environment registry key.
- The service is installed under the account `LocalSystem`, which grants it high permissions. I found this [LocalSystem](https://learn.microsoft.com/en-us/windows/win32/services/localsystem-account) doc. This doc describes that LocalSystem "is not associated with any logged-on user account", which is why system environment variables felt more appropriate.
- This is something I think is worth explicitly calling out: The environment variable docs state that system environment variables are "set for everyone", e.g. any user on a machine can access it. I had an image in my head of one windows machine = one user, but thinking about it more closely now, I'm not sure this is a valid assumption.

distributions/nrdot-collector/windows/installer.wxs
agarvin-nr: `NEW_RELIC_LICENSE_KEY` is what we call the [environment variable](https://github.com/newrelic/nrdot-collector-releases/blob/b0c6326b9cf7a12d964acf59f807fca19a564852/distributions/nrdot-collector/config.yaml#L192-L196) in the host `config.yaml`, which we package with this MSI.

Honestly, I'm not totally sold that it's desirable to prevent an install based on this specific variable. Users are able to swap out configs in the install directory, so this might be too inflexible. There are a few other options I can think of:

1. Remove this condition for more flexibility, but keep the configuration options to set default use-case config environment variables (`env:NEW_RELIC_LICENSE_KEY`, `env:OTEL_EXPORTER_OTLP_ENDPOINT` etc.)
2. Do not set environment variables via the installer and allow users to set them with commands like `Set-Itemproperty...` instead. This may be a closer match to our [other installers](https://github.com/newrelic/nrdot-collector-releases/blob/main/distributions/README.md#deb-installation).

distributions/nrdot-collector/windows/installer.wxs
agarvin-nr: V3 schema may be mandatory if we want to use msitools. According to this comment in the [msi docs](https://goreleaser.com/customization/package/msi/):

```
# msitools only supports v3.
# wixtoolset v3 supports v3.
# wix v4/v5 supports v4.
```

msitools and Wix Toolset are different services, maintained by different people. Msitools is open-source, maintained by GNOME, and lets us build MSIs on Linux. Wix/Wix toolset is owned by Firegiant and can only run on Windows. We could consider building this on a Windows runner and use Wix, but I seem to remember reading that Wix has some less-than-permissive licensing.

I think it would be worth documenting this. I can do a bit more research to validate what I wrote above, and put together a doc so we can make a more informed decision.

kb-newrelic: Thanks for the explanation, glad I asked! msitools being the only tool supported on linux combined with the licensing challenge makes total sense. Maybe we can add a comment in the second line of the xml to make it clear that this is an intentional choice as we rely on msitools.

distributions/nrdot-collector/windows/installer.wxs
kb-newrelic: I was mainly commenting on the fact that I don't understand the significance of 'Installed OR' in 'Installed OR NEW_RELIC_LICENSE_KEY'. Is this some kind of formal language in `<Condition>` blocks or does this get displayed somewhere?

Your other point is valid though. The env vars are essentially tied to the config and while we provide a default config, the customer could also supply their own making the env vars meaningless. If it is possible to provide env vars via instructions without including them in the installer, that would probably give us a cleaner separation.
If we get feedback that the UX is bad, we can always add them later.

distributions/nrdot-collector/windows/installer.wxs
kb-newrelic: Thank you for the context. A few thoughts
- Do we have any indication that the installer takes care of the second part?
> then broadcast a [WM_SETTINGCHANGE](https://learn.microsoft.com/en-us/windows/desktop/winmsg/wm-settingchange) message with lParam set to the string "Environment"
- For determining the appropriate accounts and scoping of env vars, I would suggest involving the security folks. For now, we can go with what works for e2e testing to move the other work along but there is probably a lot of nuance in this, so we should get someone that knows more about windows and its intricacies. Your last point is a good example, the default probably needs to assume a multi-user system to avoid customers accidentally sharing secrets.

One more general question: Can we make this work without the registry keys for now or does this somehow run into issues, e.g. with restarts? If that is the case, I would probably get rid of them. If not and it is to support a specific feature, e.g. restarts, we could also think about moving this into dedicated feature work 'support restarts on windows' and move on without having this implemented yet.

distributions/nrdot-collector/windows/installer.wxs
agarvin-nr: Ah, I see! It's a formal language ([Conditional Statement Syntax](https://learn.microsoft.com/en-us/windows/win32/msi/conditional-statement-syntax)) for `<Condition>` blocks. [Installed](https://learn.microsoft.com/en-us/windows/win32/msi/installed) is a special property which is set when the service is already installed on the machine; So it's basically saying "Only proceed with the install if these two properties are set".

It's been a little while since I wrote that part, but I think the intent was to allow the user to upgrade to a new version of NRDOT without being forced to supply the key (as it would already be in the registry). Thinking on it, I'm not totally sure it would work as intended; I don't know if an upgrade would try to overwrite registry keys during the install process.

I think it should be possible to supply env vars via instructions without including them in the installer; I'll test this out and see what I can find, and go that route if I'm able.

distributions/nrdot-collector/windows/installer.wxs
agarvin-nr: `Do we have any indication that the installer takes care of the second part?`

Good question! I think the answer is "yes". I just found [this document](https://docs.firegiant.com/wix3/customactions/wixsettingchange/) which states: "Windows Installer itself sends the WM_SETTINGCHANGE message for settings it changes while processing an MSI package but cannot do so for changes a package makes via custom action." I believe "custom action" refers specifically to the [Custom Action Element](https://docs.firegiant.com/wix3/xsd/wix/customaction/t), which we do not use to set registry keys. We might have to broadcast this change if we manually set registry environment variables.

`For determining the appropriate accounts and scoping of env vars, I would suggest involving the security folks. For now, we can go with what works for e2e testing to move the other work along but there is probably a lot of nuance in this...`

I'm not sure how / if the installer would function on a different account. I tried to see if contrib discussed it to see if they had any input; I found these ([Comment on initial request](https://github.com/open-telemetry/opentelemetry-collector-releases/issues/157) and [MSI design notes](https://github.com/open-telemetry/opentelemetry-collector/issues/10043#issue-2270056424)), but the only explanation I could find was that it "means no installation per user".

It might be a good idea for me to see if e2e would pass for lower permission accounts.

`One more general question: Can we make this work without the registry keys for now or does this somehow run into issues, e.g. with restarts?`

I don't know if the installer would detect shell environment variables! This is something I will test. From what I read, I'm pretty sure they would not persist on system restarts, so for that functionality would probably have to use registry keys.

distributions/nrdot-collector/windows/installer.wxs
agarvin-nr: Sure, I can add that as a comment!

.github/workflows/ci-windows.yaml
```diff
@@ -0,0 +1,219 @@
...
+          retention-days: 1
+
+  validate:
+    runs-on: windows-latest
+    needs: [build]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # required for tag metadata
```
kb-newrelic, line 117: Do we need 0 or is 1 enough?

.github/workflows/ci-windows.yaml
```diff
@@ -0,0 +1,219 @@
...
+          name: ${{ steps.find-msi.outputs.msi_artifact_name }}
+          path: ${{ steps.find-msi.outputs.msi_path }}
+          retention-days: 1
+
+  validate:
+    runs-on: windows-latest
+    needs: [build]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
kb-newrelic: Actions should be pinned at a specific commit-sha. Claude doesn't know about recent commit shas (and would need to fetch them) but you can tell it to just use the ones already used in the repo for common actions

distributions/nrdot-collector/windows/installer.wxs
```diff
@@ -0,0 +1,93 @@
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
+<!-- msitools only supports schema v3 -->
```
kb-newrelic:
```suggestion
<!-- ubuntu build machine only supports msitools which in turn only supports schema v3 -->
```

.github/workflows/ci-windows.yaml
agarvin-nr: Good catch!

.github/workflows/ci-windows.yaml
agarvin-nr: I can't think of a reason we'd need 0, I'll switch it to 1.

distributions/nrdot-collector/windows/installer.wxs
```diff
@@ -0,0 +1,93 @@
...
+                     Id="config.yaml"
+                     Name="config.yaml"
+                     Source="config.yaml"/>
+
+                  <!-- Dist runs as a service on startup -->
+                  <ServiceInstall
+                     Id="Service"
+                     Name="{{ .Binary }}"
+                     DisplayName="NRDOT Collector Host"
+                     Description="NRDOT Host Distribution of the OpenTelemetry Collector"
```
kb-newrelic:
```suggestion
                     DisplayName="NRDOT Collector"
                     Description="New Relic Distribution of OpenTelemetry Collector"
```
Removed 'Host' as it's the old distro.
I would also omit the 'the' before OTel because it's just the expansion of NRDOT which doesn't contain a 'the'.

kb-newrelic (dismissed):

distributions/nrdot-collector/windows/installer.wxs
kb-newrelic: Very helpful, thanks for the explanation!

distributions/nrdot-collector/windows/installer.wxs
kb-newrelic: As discussed, we can go ahead with contrib-parity for now and improve on this in follow-up work

kb-newrelic (approved):

PR #562, opened 2026-04-10
ci: fix lychee config

kb-newrelic:
### Summary
- Fixes [broken config](https://github.com/newrelic/nrdot-collector-releases/actions/runs/24236104302/job/70758961496#step:4:102)
- Skip steps in the workflow that expect the `.lycheecache` file. This can happen if the file wasn't in the github cache and lychee failed to run, thus not producing a cache file.


mailo-nr (approved):
LGTM

PR #564, opened 2026-04-14
test: remove tests for http/rpc metrics

kb-newrelic:
### Summary
- Follow-up to #520 which didn't delete all references to http/rpc metrics in the tests


mailo-nr (approved):
LGTM

PR #565, opened 2026-04-16
docs: update FIPS documentation

kb-newrelic:
### Summary
Update FIPS docs for GA:
- Reference correct CMVP Certificate and adjust for FIPS 140-3
- Fix manual verification instructions to also work locally on MacOS
- Adjust nmap output to show output for correct port + explain non-standard cipher names
- Add instructions for client verification
- Add workflow to ensure go version used to compile NRDOT uses FIPS-certified version of BoringSSL

### DIFF
--- a/.github/workflows/check-fips-go-version.yaml
+++ b/.github/workflows/check-fips-go-version.yaml
@@ -1,0 +1,145 @@
+name: check-fips-go-version
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '.github/workflows/ci-base.yaml'
+      - 'fips/README.md'
+      - '.github/workflows/check-fips-go-version.yaml'
+  pull_request:
+    paths:
+      - '.github/workflows/ci-base.yaml'
+      - 'fips/README.md'
+      - '.github/workflows/check-fips-go-version.yaml'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions: read-all
+
+jobs:
+  check-boringssl-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Extract Go setup config from CI
+        id: ci-go-config
+        run: |
+          # Extract go-version and check-latest from ci-base.yaml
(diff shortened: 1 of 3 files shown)


fips/validation/validate.sh
```diff
@@ -29,18 +29,18 @@ print_status() {
...
+    if [[ "$cipher" =~ ^TLS_AES_(128|256)_GCM_SHA(256|384)$ ]] ||
        [[ "$cipher" =~ ^TLS_AES_(128|256)_CCM(_8)?_SHA256$ ]]; then
-        return 0  # FIPS compliant TLS 1.3
+        return 0 
     fi
-    
+
     # TLS 1.2 FIPS-approved patterns
-    [[ "$cipher" =~ ECDHE_(ECDSA|RSA).*AES_(128|256)_GCM_SHA(256|384) ]] ||
-    [[ "$cipher" =~ DHE_RSA.*AES_(128|256)_GCM_SHA(256|384) ]] ||
```
kb-newrelic, line 42: - non-EC DHE is not FIPS compliant in the boringSSL version used by go 1.26
- Some additional key/digest sizes that are supported (GCM_SHA_384, GCM_SHA_512)

.github/workflows/check-fips-go-version.yaml
```diff
@@ -0,0 +1,145 @@
...
+  check-boringssl-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Extract Go setup config from CI
+        id: ci-go-config
+        run: |
+          # Extract go-version and check-latest from ci-base.yaml
+          GO_VERSION_SPEC=$(yq '.jobs.lint.steps[] | select(.name == "Setup Go") | .with.go-version' .github/workflows/ci-base.yaml)
```
mailo-nr, line 34: Would we be able to check the go version from the binary?  IE: `go version -m /path/to/binary`

mailo-nr (approved):
LGTM, just had a comment around the binaries golang version extraction

.github/workflows/check-fips-go-version.yaml
kb-newrelic: Looks like that would work, yeah. Given that building fips requires installing the cross compilation toolchain that would mean we'd have to run this workflow after the normal CI and download artifacts from it. We could maybe move the commit and go version into some kind of fips-spec and then have this workflow just validate that the docs match the spec and then add a test in ci-base that go version used to compile fips match the spec as well?
I'll draw something up in a follow-up.

PR #566, opened 2026-04-22
ci: validate go version used to compile fips binary

kb-newrelic:
### Summary
- Addresses [this comment](https://github.com/newrelic/nrdot-collector-releases/pull/565#discussion_r3120168911) to add golang version validation for the actual binary as suggested in [the reply](https://github.com/newrelic/nrdot-collector-releases/pull/565#discussion_r3120893281)
- `fips/spec.yaml` acts as the source of truth for the major/minor golang version and the embedded boringssl commit that we have validated that it has a FIPS certificate (see fips/README.md)
- `ci-base.yaml` validates that the binary is actually compiled using the major/minor used in the spec
- `.github/workflows/check-fips-compliance-reqs.yaml` ensures that newer patch versions of `fips/spec.yaml::go_version` still use the same commit sha for boring SSL, so that we don't have to pin the patch version as well for CI

### Note
- I thought about deriving the go version used for all setup-go steps using a script that scrapes fips/spec.yaml but discarded it again as it would pollute lots of workflows and the current setup does ensure that we keep the golang version of ci-base in line with the FIPS requirements which is the only real formal requirement we have.

### DIFF
--- a/.github/workflows/check-fips-compliance-reqs.yaml
+++ b/.github/workflows/check-fips-compliance-reqs.yaml
@@ -1,18 +1,18 @@
-name: check-fips-go-version
+name: check-fips-compliance-reqs
 
 on:
   push:
     branches:
       - main
     paths:
-      - '.github/workflows/ci-base.yaml'
+      - 'fips/spec.yaml'
       - 'fips/README.md'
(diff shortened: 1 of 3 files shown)


mailo-nr (approved):
LGTM

PR #582, opened 2026-05-11
chore: migrate windows ci to ec2

agarvin-nr:
#### Terraform Refactor

The old approach would spin up all EC2s at once with `terraform apply` in a single scenario. Unfortunately, the New Relic E2E Integration action can only generate and query against a single test key per scenario. This meant that the same test key attribute was assigned to all platforms, so they were not being tested separately. This PR refactors terraform and our test specs:

- Adds inputs for platform and version to nightly terraform, allowing a single scenario to spin up a single EC2 to test against
- Creates modules to separate platform (windows and ubuntu) logic
- Creates a reusable "common" module to contain platform-agnostic infrastructure (S3 access, network infrastructure, etc.)
- Creates separate e2e scenarios for platforms / versions (Ubuntu 22.04 and 24.04, Windows Server 2025)

#### Windows CI Migration to EC2

- Adds terraform setup and teardown steps to the Windows test spec file (using a separate windows test spec file was necessary to avoid testing windows on fips)
- Moves GitHub Actions windows test logic to EC2
- Removes old windows CI

### Validation

- Manual execution of [nightly workflow](https://github.com/newrelic/nrdot-collector-releases/actions/runs/26127496943) passes

### Potential Improvements

At the moment, each test spec runs each scenario in sequence. As each spec tests at most two versions, this is not too time consuming, but once we expand our tests to accommodate additional versions (e.g. windows 2016, 2019, 2022, 2025), we might want to find a way to run these in parallel. I wasn't able to find a way to get the e2e action to selectively run scenarios in a single file, so we may need to generate individual spec files for supported versions.


.github/workflows/deploy-and-e2e.yaml
```diff
@@ -207,4 +217,50 @@ jobs:
...
+        uses: ./.github/actions/terraform-setup
+        with:
+          terraform_version: 1.9.8
+          aws_account_id: ${{ secrets.aws_account_id }}
+          working_directory: ./test/terraform/nightly
+          workspace: ${{ needs.setup.outputs.workspace }}-windows
+          assume_role: "false"
+      
+      - name: Run Action-based E2E Tests
+        uses: newrelic/newrelic-integration-e2e-action@c548a33a0c2941a4db4899ded766923eb3178e0e # v1
```
mailo-nr: we should use the same sha as above or update both

.github/workflows/deploy-and-e2e.yaml
```diff
@@ -207,4 +217,50 @@ jobs:
           account_id: ${{ secrets.nr_account_id }}
           api_key: ${{ secrets.nr_api_key }}
           license_key: ${{ secrets.nr_ingest_key }}
+          region: ${{ vars.NR_REGION }}
+
+  e2e-tests-windows:
```
mailo-nr, line 277: If we're splitting these would it make sense to make it more symmetric?  Something like: 

```
 e2e-kind:
    steps:
      - configure kubectl
      - run e2e action → spec-kind.yaml  # pure assertions

  e2e-ec2-ubuntu:
    strategy:
      matrix:
        version: ["22.04", "24.04"]
    steps:
      - terraform apply -var="platform=ubuntu" -var="platform_version=${{ matrix.version }}"
      - run e2e action → spec-ubuntu.yaml  # pure assertions
      - terraform destroy  # if: always()

  e2e-ec2-windows:
    if: ${{ !inputs.fips }}
    strategy:
      matrix:
        version: ["2025"]
    steps:
      - terraform apply -var="platform=windows" -var="platform_version=${{ matrix.version }}"
      - run e2e action → spec-windows.yaml  # pure assertions
      - terraform destroy  # if: always()

File: test/terraform/modules/ec2/windows/main.tf [hunk 3]
```diff
@@ -0,0 +1,105 @@
...
+  subnet_id = module.common_infrastructure.private_subnet_ids[0]
+  vpc_security_group_ids = [module.common_infrastructure.security_group_id]
+  iam_instance_profile = module.common_infrastructure.instance_profile_name
+
+  tags = {
+    Name = "${var.test_environment}-${var.collector_distro}-${local.instance_identifier}"
+  }
+
+  user_data_replace_on_change = true
+  user_data                   = <<-EOF
```
mailo-nr, line 43: not a blocker but this might be a good candidate for https://developer.hashicorp.com/terraform/language/functions/templatefile

.github/workflows/deploy-and-e2e.yaml
```diff
@@ -198,7 +209,6 @@ jobs:
           LICENSE_KEY: ${{ secrets.nr_ingest_key }}
           IMAGE_TAG: ${{ needs.setup.outputs.image-tag }}
           IMAGE_REPO: ${{ secrets.aws_account_id }}.dkr.ecr.us-east-1.amazonaws.com/${{ inputs.distro }}
-          CLUSTER_NAME: aws-ci-e2etest
```
mailo-nr, line 201: Assuming this is set automatically now?

distributions/nrdot-collector/test/spec-nightly-windows.yaml
```diff
@@ -106,10 +106,20 @@ description: nrdot-collector E2E Test (Windows)
         lowerBoundedValue: 1
 
 scenarios:
-  - description: host telemetry
+  - description: windows 2025 - ec2 host metrics
+    before:
+      - | 
+        cd ../../../test/terraform/nightly &&
+        terraform apply -auto-approve \
```
mailo-nr, line 113: Not for this pr since it's existing already but we may want to look into moving TF out of the spec files since it higher level infra and we already run tf in the e2e workflow, we'd basically be hoisting this up into the parent workflow

.github/workflows/deploy-and-e2e.yaml
agarvin-nr: Good catch! I'll use the most recent `d0c7c2b` hash (which is an update I pushed to their repo to fix a windows path issue)

.github/workflows/deploy-and-e2e.yaml
agarvin-nr: I see that we use this name to update the aws kubeconfig up above, but I wasn't able to find a place where  this CLUSTER_NAME variable was used in the test spec.

distributions/nrdot-collector/test/spec-nightly-windows.yaml
agarvin-nr: I really like this idea! As you touched on in your other comment, moving them out of spec files would make it a lot nicer to test different versions (especially since they could run in parallel with github actions matrix). If not for this PR, then I think we should make it an AC for supporting additional versions of windows.

.github/workflows/deploy-and-e2e.yaml
agarvin-nr: I agree 100%! I'll go ahead and split them as a part of this PR. Moving the terraform files will be a little bit more tricky. New relic integration e2e automatically generates its own scenario keys, which we pass into the EC2 during the setup phase with `-var=testKey={SCENARIO_TAG}`. We would need to generate our own scenario tag and pass it in via the action's [scenario tag parameter](https://github.com/newrelic/newrelic-integration-e2e-action#usage), but we can probably take inspiration from [their implementation](https://github.com/newrelic/newrelic-integration-e2e-action/blob/d0c7c2b8f9af227365d71c6b7fddbc024c9e331a/internal/runtime/runner.go#L161) to generate it.

mailo-nr (approved):
LGTM

PR #583, opened 2026-05-12
chore: search through all artifacts while verifying artifacts in nightly

agarvin-nr:
### Summary
Nightly only searches through the first 30 artifacts when verifying artifacts. This has become a problem with the addition of the Windows CI workflow, which creates additional artifacts and has thus pushed the nightly artifact outside of the first 30. This results in [nightly failures on the main branch](https://github.com/newrelic/nrdot-collector-releases/actions/runs/25765694189/job/75677567235#step:3:41).

This PR fixes this by searching the entire list of artifacts instead of only the first 30.

### Validation
[Nightly Passes](https://github.com/newrelic/nrdot-collector-releases/actions/runs/25767602427)

### DIFF
--- a/.github/actions/verify-artifacts/action.yaml
+++ b/.github/actions/verify-artifacts/action.yaml
@@ -44,7 +44,7 @@ runs:
         echo "Verifying $BUILD_TYPE artifacts exist for ${{ inputs.distribution }} with SHA: ${{ inputs.commit-sha }}"
         echo "Looking for artifact: $ARTIFACT_NAME"
 
-        if gh api "/repos/${{ github.repository }}/actions/artifacts" \
+        if gh api --paginate "/repos/${{ github.repository }}/actions/artifacts" \
            --jq ".artifacts[] | select(.name == \"$ARTIFACT_NAME\")" \
            | grep -q .; then
           echo "found=true" >> $GITHUB_OUTPUT


kb-newrelic (approved):

.github/actions/verify-artifacts/action.yaml
```diff
@@ -44,7 +44,7 @@ runs:
         echo "Verifying $BUILD_TYPE artifacts exist for ${{ inputs.distribution }} with SHA: ${{ inputs.commit-sha }}"
         echo "Looking for artifact: $ARTIFACT_NAME"
 
-        if gh api "/repos/${{ github.repository }}/actions/artifacts" \
+        if gh api --paginate "/repos/${{ github.repository }}/actions/artifacts" \
```
kb-newrelic, line 47: In case this fails, it might have to do with a different structure of the response but I assume this should work because the jq option should be aware of this and grep doesn't care about json.

> In --paginate mode, all pages of results will sequentially be requested until there are no more pages of results. For GraphQL requests, this requires that the original query accepts an $endCursor: String variable and that it fetches the pageInfo{ hasNextPage, endCursor } set of fields from a collection. Each page is a separate JSON array or object. Pass --slurp to wrap all pages of JSON arrays or objects into an outer JSON array.

agarvin-nr: That's a good thought! I think it should work (running nightly manually seems to), but that `--slurp` note seems like it would come in handy to fix it if need be.

PR #588, opened 2026-05-28
feat: publish msi and update readme

agarvin-nr:
This should not be merged until the MSI is ready for publishing.

### Summary

- Enables MSI publishing
- Adds install and troubleshooting info to readmes
- Adds a note about possible windows delays caused by the upstream tier 2 support policy for Windows

### DIFF
--- a/distributions/README.md
+++ b/distributions/README.md
@@ -115,6 +115,42 @@ tar -xzf collector.tar.gz
 NEW_RELIC_LICENSE_KEY="${license_key}" ./nrdot-collector --config ./config.yaml 
 ```
 
+### Windows MSI and Archives
+All windows install options are available under [Releases](https://github.com/newrelic/nrdot-collector-releases/releases). Windows binaries and MSI files are only provided for `amd64` architecture.
+
+#### MSI Installation
+NRDOT must be installed from an Administrator account, and will run as an automatic service under the `Local System` service account.
+
+```powershell
+$collector_distro = "nrdot-collector"
+$collector_version = "1.15.1"
+$license_key = "YOUR_LICENSE_KEY"
+Invoke-WebRequest -Uri "https://github.com/newrelic/nrdot-collector-releases/releases/download/$collector_version/$collector_distro.msi" -OutFile "nrdot-collector.msi"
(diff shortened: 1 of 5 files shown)


distributions/README.md
```diff
@@ -115,6 +115,44 @@ tar -xzf collector.tar.gz
...
+$license_key = "YOUR_LICENSE_KEY"
+Invoke-WebRequest -Uri "https://github.com/newrelic/nrdot-collector-releases/releases/download/$collector_version/$collector_distro.msi" -OutFile "nrdot-collector.msi"
+Start-Process -Wait -PassThru msiexec.exe -ArgumentList "/i nrdot-collector.msi /qn"
+New-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\$collector_distro" `
+                -Name 'Environment' `
+                -PropertyType MultiString `
+                -Value @(
+                  "NEW_RELIC_LICENSE_KEY=$license_key"
+                  # Add any other config environment variables to this registry key (comma-separated)
+                ) `
```
kb-newrelic: Does this work as-is? I can't verify but some research I did makes me wonder if the comment inside the array wouldn't eat the closing bracket? Wouldn't surprise me as bash parsing works similarly and makes placing comments always a pain in scripts.
Gemini suggests something like
```
$RegistryArgs = @{
    Path         = "HKLM:\SYSTEM\CurrentControlSet\Services\$collector_distro"
    Name         = 'Environment'
    PropertyType = 'MultiString'
    Value        = @(
        "NEW_RELIC_LICENSE_KEY=$license_key"
        # Add any other config environment variables to this registry key (comma-separated)
    )
    Force        = $true
}

New-ItemProperty @RegistryArgs
```

kb-newrelic (dismissed):

distributions/README.md
agarvin-nr: I just added this comment to the nightly test and tested both approaches - They both seemed to work, but I think this pattern is easier to read so I went with this suggestion. [Nightly passes](https://github.com/newrelic/nrdot-collector-releases/actions/runs/26913771188/job/79398553059).

kb-newrelic (approved):

PR #589, opened 2026-05-29
refactor: release windows artifacts separately

agarvin-nr:
WIP

### Summary
* Update CI to release windows artifacts separately
* Some fixes for our act workflows (two of our GHA inputs were recently moved from secrets to vars, but .secrets.example still listed them as secrets)


agarvin-nr:
Closing this: This work has been deprioritized (and is likely unnecessary). There is probably a better approach if it gets picked up again.

PR #591, opened 2026-06-01
chore: fix act variables and add act-generated files to .gitignore

agarvin-nr:
### Summary

* Creates a file `.actvars` to define variables (Some of our GitHub variables were recently moved from secrets to vars. Act will not try to grab regular variables from the `.secrets` file, so they needed to be defined elsewhere.)
* Adds `.cache` directory to .gitignore (This is created by trivy when running in act)
* Only prepare goreleaser artifacts on input.publish (Prevents certain artifacts like `docker_manifest_sha` being created during act workflow runs)
* Adds some missing secrets to `.secrets.example`


Makefile.dev
```diff
@@ -6,22 +6,26 @@ ci_custom_matrix:
 	@# repeat --matrix arg for multiple distros
 	act push -W .github/workflows/ci.yaml \
 		--matrix distribution:nrdot-collector \
-		--matrix fips:false
+		--var-file .actvars
```
kb-newrelic: Can this not live in the `.actrc` instead? That would make it more robust if someone writes a manual act command

kb-newrelic (approved):

PR #597, opened 2026-06-09
ci: move terraform setup and destroy to workflows

agarvin-nr:
### Summary
Followup PR to address [feedback](https://github.com/newrelic/nrdot-collector-releases/pull/582#discussion_r3282919361) in the initial MSI PR. This moves the terraform apply / destroy steps out of the spec file and into the parent workflow. This has the following benefits:

* Faster nightly runs from different versions on parallel runners (previously would run all platform versions in sequence)
* Reduces copy-paste for testing additional versions in nightly (now we only need to add additional matrix inputs)

### Validation
* [Nightly passes](https://github.com/newrelic/nrdot-collector-releases/actions/runs/27243649310)
* Terraform correctly destroys on workflow failure (see: [test run](https://github.com/newrelic/nrdot-collector-releases/actions/runs/27243385984/job/80452120417#step:7:1) with added failure step)

### DIFF
--- a/.github/workflows/deploy-and-e2e.yaml
+++ b/.github/workflows/deploy-and-e2e.yaml
@@ -314,8 +314,8 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
       - name: Configure AWS Credentials
-        id: configure-aws-credentials-for-eks
-        uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6
+        id: configure-aws-credentials
+        uses: aws-actions/configure-aws-credentials@d979d5b3a71173a29b74b5b88418bfda9437d885 # v6
         with:
           role-to-assume: ${{ secrets.aws_oidc_role_arn }}
           aws-region: us-east-1
--- a/.github/workflows/deploy-and-e2e.yaml
+++ b/.github/workflows/deploy-and-e2e.yaml
@@ -260,7 +260,7 @@ jobs:
         shell: bash
         run: |
           RANDOM_SUFFIX=$(echo $RANDOM$RANDOM | md5sum | cut -c1-7)
-          echo "SCENARIO_TAG=e2e-${GITHUB_SHA:0:7}-${RANDOM_SUFFIX}" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-and-e2e.yaml
```diff
@@ -284,8 +314,8 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
       - name: Configure AWS Credentials
-        id: configure-aws-credentials-for-eks
-        uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6
+        id: configure-aws-credentials
+        uses: aws-actions/configure-aws-credentials@d979d5b3a71173a29b74b5b88418bfda9437d885 # v6
```
agarvin-nr, line 318: Updated this hash to match all the other ones in the repo

.github/workflows/deploy-and-e2e.yaml
```diff
@@ -250,8 +253,22 @@ jobs:
...
           working_directory: ./test/terraform/nightly
-          workspace: ${{ needs.setup.outputs.workspace }}-ubuntu
+          workspace: ${{ needs.setup.outputs.workspace }}-ubuntu-${{ matrix.version }}
           assume_role: "false"
+      
+      - name: Generate E2E scenario tag
+        shell: bash
+        run: |
+          RANDOM_SUFFIX=$(echo $RANDOM$RANDOM | md5sum | cut -c1-7)
+          echo "SCENARIO_TAG=e2e-${GITHUB_SHA:0:7}-${RANDOM_SUFFIX}" >> $GITHUB_ENV
```
mailo-nr: Does the version get added to this tag further down the chain?  If not it might be helpful.

mailo-nr (dismissed):
LGTM, just the comment about the tag, but that's a nit.

.github/workflows/deploy-and-e2e.yaml
agarvin-nr: The version does not get added to the tag down the line - This just follows a similar pattern as [newrelic e2e action](https://github.com/newrelic/newrelic-integration-e2e-action/blob/d0c7c2b8f9af227365d71c6b7fddbc024c9e331a/internal/runtime/runner.go#L161) to generate a random tag to query against. I can see how adding platform and version to the tag could be helpful for debugging in NRQL though, so I'll go ahead and do that!

mailo-nr (approved):
LGTM

PR #596, opened 2026-06-09
feat(otel-receiver): Add otel receiver for ohai otel instrumentation

vagrawal-newrelic:
Add Redis, HAProxy, ECS, and Kafka receivers to nrdot-collector-experimental for testing prior to promotion to the release distribution.           
                                                                                                                                                     
  Receivers added (experimental only):                                                                                                               
  - redisreceiver (OHI-redis) — https://docs.newrelic.com/docs/opentelemetry/integrations/redis/overview/                                            
  - kafkareceiver (OHI-kafka) — https://docs.newrelic.com/docs/opentelemetry/integrations/kafka/overview/
  - haproxyreceiver (OHI-haproxy) — https://docs.newrelic.com/docs/opentelemetry/integrations/haproxy/overview/
  - awsecscontainermetricsreceiver (OHI-ecs) — https://docs.newrelic.com/docs/opentelemetry/integrations/ecs-monitoring/overview/

  Other changes:
  - Updated component-inventory.md with sub-categories (OHI-redis, OHI-kafka, OHI-haproxy, OHI-ecs, OHI-elasticsearch, OHI-nginx, OHI-rabbitmq)
  - Updated README.md with documentation links for all OHI integrations
  - Added temporary thrift replace directive in experimental to fix CVE-2026-41602 (can be removed once contrib is bumped to v0.151.0+)
  - Regenerated THIRD_PARTY_NOTICES.md

  Process:
 - [x] Receivers added to **experimental only** (no changes to `nrdot-collector`)                                                                   
  - [ ] Test with experimental release and verify end-to-end                                                                                         
  - [ ] Confirm working → request promotion to `nrdot-collector` 

These receivers unblock the guided install flows and OIL recipes for Redis, HAProxy,
and Apache that currently fail with unknown type errors when using NRDOT.

PR #599, opened 2026-06-11
feat(experimental): add additional ohi components

kb-newrelic:
### Summary
- Add components suggested in #596 to experimental
- We'll keep the other PR as reference for promotion to prod as it already has the links and component-inventory attribution but we should only add those when the promotion happens and will also need to add the links either to an exception list for the link checker CI or add them in a follow-up PR


kb-newrelic:
Blocked by bump to new component versions, will rebase once new versions are incorporated into NRDOT

agarvin-nr (approved):
LGTM!

PR #609, opened 2026-06-22
feat: add ohi components to nrdot-collector

kb-newrelic:
### Summary
- Add additional OHI components to core after testing via experimental #599 was successful


mailo-nr (approved):
LGTM

PR #614, opened 2026-06-30
chore: update bump-component-versions and nrdot-collector-builder with ability to update nr forks repo

agarvin-nr:
### Summary
* Updates `nrdot-collector-builder` and `bump-component-versions.sh` to detect and update manifest.yaml components from the `newrelic-forks/opentelemetry-collector-contrib` repository.
* Runs `bump-collector-versions.sh` - This only alphabetizes the manifests, and should supercede [otelcomm-bot's current pr](https://github.com/newrelic/nrdot-collector-releases/pull/603/changes)
* Fixes an issue with NRDOT-Upstream component version validation. Currently, the `nrdot-collector-builder` allows allows nrdot to have a greater minor version than upstream, and disallows upstream from having a patch version greater than nrdot. The expected behavior is that these components have matching major and minor versions, with no restrictions on patch versions. This PR fixes this by changing the check from  from `nrdot >= upstream` to `majorMinor(nrdot) == majorMinor(upstream)`.

### Validation
* Adds additional unit tests and test config to `nrdot-collector-builder` to test new functionality. Note: Tests [pass](https://github.com/newrelic/nrdot-collector-releases/actions/runs/28545069836/job/84628204142#step:5:9), but CI is [intentionally](https://github.com/newrelic/nrdot-collector-releases/pull/309) set to not block on these failing.


cmd/nrdot-collector-builder/cmd/manifest/update_test.go
```diff
@@ -82,3 +82,50 @@ func TestUpdateCmd_RunE_InvalidConfig(t *testing.T) {
...
+	assert.NoError(t, err)
+	defer os.Remove(tempFile.Name()) // Clean up the file after the test
+
+	// Write the loaded YAML data to the temporary file
+	_, err = tempFile.Write(yamlData)
+	assert.NoError(t, err)
+	tempFile.Close() // Close the file to ensure the changes are flushed
+
+	targetStable := "v1.61.0"
+	targetBeta := "v0.155.0"
```
agarvin-nr, line 103: This test passes even though v0.155.0 is not yet released for the components repo and nr fork repo. This is because `nrdot-collector-builder` does not seem to validate that pinned versions exist. We validate version existence in `bump-component-versions.sh`, but if we work on this program again we might want to add that check.

cmd/nrdot-collector-builder/internal/manifest/config_test.go
```diff
@@ -177,10 +207,79 @@ func TestIsCompatibleWithNrdotComponent(t *testing.T) {
...
-			result := isCompatibleWithNrdotComponent(tt.nrdotVersion, tt.betaVersion)
+			result := isCompatibleWithNrComponent(tt.nrdotVersion, tt.betaVersion)
 			assert.Equal(t, tt.expectedMatch, result,
 				"isCompatibleWithNrdotComponent(%s, %s) = %v, want %v",
 				tt.nrdotVersion, tt.betaVersion, result, tt.expectedMatch)
 		})
 	}
 }
+
+func TestIsCompatibleWithNrVersions(t *testing.T) {
```
kb-newrelic, line 218: all tests have the patch version aligns as well, it would probably be good to add test cases where they don't align, e.g.  0.154.4 fork is compatible with 0.154.0 beta

kb-newrelic (dismissed):

cmd/nrdot-collector-builder/internal/manifest/config_test.go
```diff
@@ -272,6 +272,27 @@ func TestIsCompatibleWithNrVersions(t *testing.T) {
...
+			nrForkContribVersion: "v0.142.4",
+			betaVersion:          "v0.142.0",
+			expectedMatch:        true,
+		},
+		{
+			name:                 "beta patch version higher",
+			nrdotVersion:         "v0.142.0",
+			nrForkContribVersion: "v0.142.0",
+			betaVersion:          "v0.142.5",
+			expectedMatch:        false,
```
kb-newrelic: a higher beta patch version shouldn't lead to an invalid match. The patch version should essentially be ignored for matching afaik because we can't infer anything about the beta patch version used by nrdot based on its version, it's only guaranteeing to align the minor, i.e. 0.142.1 could use beta 0.140.0 or 0.140.2, there is no restriction.

kb-newrelic (approved):

PR #615, opened 2026-07-06
docs: fix mssql link

kb-newrelic:
### Summary
- Fixes failing [link check](https://github.com/newrelic/nrdot-collector-releases/actions/runs/28810973382) due to docs reorg

PR #616, opened 2026-07-07
ci: add chloggen and workflows for changelog generation

agarvin-nr:
### Summary
Adds OpenTelemetry's `chloggen` tool to this repo for changelog management.

* Adds chloggen make targets
* Enforces a changelog entry is created with each PR
* Changelog is skippable with:
    * "Skip Changelog" label
    * Any conventional commit prefix `chore*`, `build*`, `ci*`, `refactor*`, `style*`, or `test*`

#### Release Flow

1. Developers create a changelog entry with every PR if needed
2. Bumping component versions with `ci-bump-component-versions` workflow automatically creates changelog entry like `Bump component versions from v... to v...`
    a. If relevant changes are found with the `version-bump-analysis` skill, we can either elaborate using the `subtext:` field on the autogenerated entry, or generate our own entries as needed.
3. Bumping nrdot version with `release-bump-version-pr` will consolidate all changelog entries using `make chlog-update` into `CHANGELOG.md`
4. The `release-draft` workflow extracts the most recent changes from `CHANGELOG.md` and provides to goreleaser via `--release-notes` (this replaces the previous changelog generation)
5. Unchanged from before: After we publish, the `release-publish` workflow will issue a docs PR from `github.event.release.body`

#### Note: These are the updated outputs, having used `chloggen-wrapper.sh`
Added changelog entry for PR #615 (which will be included next release) - `make chlog-validate` passes

Tested script outputs:

* `scripts/generate-version-bump-chlog.sh` (used in `ci-bump-component-versions.yaml`) with fake upstream version `v0.999.0`:
```agarvin_chloggen.yaml
change_type: feature
component: distributions
issues: [616]
note: Bump otel component versions from v0.154.0 to v0.999.0
subtext:

```

* Tested changelog.md output of `release-bump-version-pr` in Act (using fake version + both chlog entries above)
```CHANGELOG.md
<!-- This file is autogenerated. See CONTRIBUTING.md for instructions to add an entry. -->

<!-- next version -->

## 1.90.0

### Features

- `distributions`: Bump otel component versions from v0.154.0 to v0.999.0 (#616)

### Docs

- `distributions`: Fix broken mssql link (#615)

<!-- previous-version -->

```

* Tested output of [release notes extraction](https://github.com/newrelic/nrdot-collector-releases/pull/616/changes#diff-5d2f218e423fbc2f2042d2520fab4524acaab776ce1fc363908d18d312a06459R72-R88) script (from above changelog):
```release-notes.md
## 1.90.0

### Features

- `distributions`: Bump otel component versions from v0.154.0 to v0.999.0 (#616)

### Docs

- `distributions`: Fix broken mssql link (#615)

```


.chloggen/config.yaml
```diff
@@ -0,0 +1,13 @@
...
+    - user
+entries_dir: .chloggen
+template_yaml: .chloggen/TEMPLATE.yaml
+summary_template: .chloggen/summary.tmpl
+components:
+    - distributions
+    - distributions/nrdot-collector
+    - distributions/nrdot-collector-experimental
+    - docs
+    - examples
```
agarvin-nr: Adding this list to the config enforces that only these components may be added to each entry. If this is too restrictive, we could instead omit this section entirely (then users would still be mandated to add some "component" to their changelog, but would be free to choose whatever.)

scripts/validate-actions-hashes.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
agarvin-nr, line 1: I've had a few near-misses (from copy-paste or Claude) with GitHub Actions commit hashes being pinned to versions differing from the rest of the repo. I made this script to try and catch that more reliably, and I thought it might be nice for CI.

Examples of it working: [Failing run from this PR](https://github.com/newrelic/nrdot-collector-releases/actions/runs/29051945528/job/86234577038#step:6:9) and [successful run after I fixed it](https://github.com/newrelic/nrdot-collector-releases/actions/runs/29052308246/job/86235757283#step:6:9)

.github/workflows/ci-bump-component-versions.yaml
```diff
@@ -134,11 +135,32 @@ jobs:
               
               echo "Updated PR #$pr_number"
             else
-              echo "Creating new PR..."
+              echo "Creating new draft PR..."
```
agarvin-nr, line 177: `chloggen` mandates that there is at least one value entered into the `issues` field of each entry, which is meant to represent either an issue number (which we don't use) or pull request number.

For our auto-generated PRs, we cannot generate a passing changelog before we have a PR number, which means we must create a PR beforehand. This leads to the following issue:

1. This workflow opens a PR
2. CI runs against initial PR and fails
3. This workflow adds changelog entry
4. CI runs against workflow and passes

The CI run in step 2 will always fail, either because of the lack of changelog, or because it was cancelled by the CI run at 4. I don't think this would affect the release process, but my idea here was to mark it as a draft PR here, then mark it as 'ready' after the changelog is added so we don't get emailed until it's ready.

.chloggen/kbauer_update_use_case_docs.yaml
```diff
@@ -0,0 +1,14 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
```
kb-newrelic: It would be ideal if we had control over the categories and could use our existing labels like '[Features](https://github.com/newrelic/docs-website/pull/24615/changes#diff-8283aee99c3d5b203dfad719504e82a2ae250a989d85a84ec6ab2771fa61fa42R10)'. Eventually we'll also want to inject the changes into the metadata like we do for the [host use case](https://github.com/newrelic/docs-website/pull/24720/changes#diff-f5c26fa2ab83dd8e5b83a59cfc42e32cc21cbdad97f5116ef2278551febd530bR9-R11).

I think it's worth seeing if we can extend and/or configure chloggen somehow to adjust this behavior. One solution might also be wrapping it and have a postprocessor adjust the output?
If it gets too complicated, we can skip it for now, but a plan how we can get to that state would be great.

kb-newrelic: If the template is essentially fixed to use the terminology OTel has established, than we can probably live with just relabelling it in the changelog template, would just be ideal if we don't have that translation layer, so that the chloggen files already use our terminology.

.github/workflows/changelog.yaml
```diff
@@ -0,0 +1,116 @@
...
+    branches:
+      - main
+
+permissions:
+  contents: read
+
+env:
+  # Make sure to exit early if cache segment download times out after 2 minutes.
+  # We limit cache download as a whole to 5 minutes.
+  SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2
```
kb-newrelic: curious what this is about? Is there some kind of link validation in chloggen?

.github/workflows/changelog.yaml
```diff
@@ -0,0 +1,116 @@
...
+            echo "See CONTRIBUTING.md for more details."
+            echo "Alternately, add \"chore:\" to the title of the pull request or add the \"Skip Changelog\" label if this job should be skipped."
+            false
+          else
+            echo "A changelog entry was added to the ./.chloggen/ directory."
+          fi
+
+      - name: Validate ./.chloggen/*.yaml changes
+        run: |
+          make install-tools
```
kb-newrelic: chlog-validate should have install chloggen as a dependency and not require make install-tools beforehand or am I missing something?

.github/workflows/release-bump-version-pr.yaml
```diff
@@ -39,7 +46,14 @@ jobs:
           echo "old_version=${old_version}" >> $GITHUB_OUTPUT
           echo "branch=nrdot-release/${{ inputs.new_version }}" >> $GITHUB_OUTPUT
 
+      - name: Generate changelog
+        id: chloggen
+        run: |
+          make install-tools
```
kb-newrelic: same as above, chlog-update should have install chloggen as a dependency

internal/tools/go.mod
```diff
@@ -12,8 +10,13 @@ require (
 require (
 	github.com/cyphar/filepath-securejoin v0.4.0 // indirect
 	github.com/google/licenseclassifier v0.0.0-20250108172508-0c3884c4cd43 // indirect
+	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/sergi/go-diff v1.3.1 // indirect
+	github.com/spf13/cobra v1.10.2 // indirect
+	github.com/spf13/pflag v1.0.10 // indirect
+	go.opentelemetry.io/build-tools/chloggen v0.30.0 // indirect
```
kb-newrelic: should probably be a direct dependency now?

kb-newrelic (changes requested):

internal/tools/go.mod
agarvin-nr: Good catch! I forgot to `go mod tidy` this after adding chloggen, I'll do that.

.github/workflows/changelog.yaml
agarvin-nr: You are correct, it shouldn't require make install-tools, I will remove it. A few of the other new make targets were missing the install dependency too, so I'll add those in.

.github/workflows/changelog.yaml
agarvin-nr: That's a great question - I took most of this workflow from the components repo and missed this. I just did some research and found that it's an environment variable [used to set cache download timeout](https://github.com/actions/cache#v3) in `actions/cache`. In this instance, it seems it's used to configure `actions/setup-go` (which [uses](https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs) the same cache mechanism via [toolkit/cache](https://github.com/actions/toolkit/tree/main/packages/cache#actionscache).)

The default timeout is 10 minutes. I imagine this would be a lot more relevant for repos with bigger caches than releases (like our components repo or upstream contrib), but we're probably safe to remove this from here if need be.

.chloggen/kbauer_update_use_case_docs.yaml
agarvin-nr: I had the same thought about trying to control these categories. I briefly looked at trying to configure these values, and unfortunately it looks like the change types are [hard-coded](https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/6386a329ce41d21f0cd3b144d67c6a60a6797222/chloggen/internal/chlog/entry.go#L32).

I agree that it would be much nicer to have the actual section names reflected in the changelog entries though. I can look into adding an intermediary processor that sits in between the changelog .yamls and chloggen.

scripts/chloggen-wrapper.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/bash
+# Copyright New Relic, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# This script serves as a wrapper around chloggen so that we can use our own change types.
+# Chloggen's default change types are "breaking", "deprecation", "bug_fix", "new_component", and "enhancement".
+# These values are hard-coded and not configurable, necessitating a wrapper if we want to use our own custom types.
+
```
kb-newrelic, line 8:
```suggestion

set -euo pipefail
```

scripts/validate-actions-hashes.sh
```diff
@@ -0,0 +1,28 @@
...
+# Copyright New Relic, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
```
kb-newrelic:
```suggestion
set -euo pipefail
```

scripts/generate-version-bump-chlog.sh
```diff
@@ -0,0 +1,45 @@
...
+# Generates a changelog entry for an otel component version bump.
+#
+# Must be run after the PR is created so that `make chlog-new` can populate the
+# .issues field with the PR number.
+#
+# Usage: generate-version-bump-chlog.sh -c <current_beta_core> -n <next_beta_core>
+#   -c  current otel beta core version (e.g. v0.147.0)
+#   -n  next otel beta core version    (e.g. v0.148.0)
+
+set -e
```
kb-newrelic:
```suggestion
set -euo pipefail
```

CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+<!-- This file is autogenerated. See CONTRIBUTING.md for instructions to add an entry. -->
+
+<!-- next version -->
```
kb-newrelic, line 3: To make the transition a bit more intuitive, I'd suggest popoluating a 1.21.0 entry by hand and adding a note at the bottom that all previous changelogs can be found in the respective github releases.

.chloggen/README.md
```diff
@@ -0,0 +1,33 @@
+### Changelog folder
+
+This repo uses `chloggen` to manage its changelog files. You can find the source code for the tool [here](https://github.com/open-telemetry/opentelemetry-go-build-tools/tree/main/chloggen).
```
kb-newrelic: Probably worth linking to the section in CONTRIBUTING.md to explain the wrapper and usage

.chloggen/README.md
```diff
@@ -0,0 +1,33 @@
...
+# The CHANGELOG file or files to which 'chloggen update' will write new entries
+# (Optional) Default filename: CHANGELOG.md
+change_logs:
+  user: CHANGELOG.md
+
+# The default change_log or change_logs to which an entry should be added.
+# If 'change_logs' is specified in this file, and no value is specified for 'default_change_logs',
+# then 'change_logs' MUST be specified in every entry file.
+default_change_logs: [user]
+```
```
kb-newrelic, line 34: Would move that directly into `config.yaml`. Someone editing that might not look in here

agarvin-nr: That's a good idea, done!

CHANGELOG.md
agarvin-nr: I just did so! I have the entry matching the current release.

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
 <!-- This file is autogenerated. See CONTRIBUTING.md for instructions to add an entry. -->
 
 <!-- next version -->
+
+## 1.21.0
+
+### Features
+
+- `distributions`: feat: Bump otel component versions from v0.154.0 to v0.156.0 (#616)
+- `distributions`: feat: Bump `nroracledbreceiver` and `nrsqlserverreceiver` to v0.156.1 (#616)
```
kb-newrelic: As discussed, lets remove the prefix from the changelog if possible and stick to what we've done before until we have other requirements. If chloggen does not support empty `component`, then we can skip with the current setup but let's rename components for the sake of brevity:

`distributions` -> `general`
`distributions/nrdot-collector` -> `nrdot-collector`
`distributions/nrdot-collector-experimental` -> `nrdot-collector-experimental`

kb-newrelic (changes requested):

kb-newrelic (approved):

PR #619, opened 2026-07-15
ci: automatically check and report nr-forks drift

agarvin-nr:
### Summary

- Fixes an issue where the workflow would incorrectly fail with an error on non-matching NR component versions (e.g.`nrdot-collector-components` at v0.155.0, nr forksat v0.154.0) 
- Adds an automated check for version drift. If the drift exceeds some grace period (currently 14 days), it will send a slack message to our bot channel. Example output:

> :warning: **Components repo and Forks repo versions have been out-of-sync for over 14 days!**
>
> nrdot-collector-components version: `v0.156.0`
> newrelic-forks/opentelemetry-collector-contrib version: `v154.6`
> Days out-of-sync: **14**

- Re-architects component workflow; Adds a read-only script to check versions and calculate drift, which provides information to the slack message and component bumping workflow. This was done because the original would have different-shaped jsons for successes (generated by nrdot-collector-builder) than for failures (drift time).

### Validation
Workflow [skips without error](https://github.com/newrelic/nrdot-collector-releases/actions/runs/29436106646/job/87423056525#step:7:32) now


mailo-nr (approved):
LGTM

PR #623, opened 2026-07-21
ci: remove ci-host-config-drift workflow

kb-newrelic:
## Summary
- Removes the `ci-host-config-drift` GitHub workflow (Agent Control config drift check) which has been moved to an internal use case repo
- Removes `distributions/nrdot-collector/Makefile`'s `assert-config-invariants` target and its backing script `assert-invariants-agent-control.sh`, since both were exclusively used by the removed workflow and are unreferenced elsewhere

## Test plan
- [x] Confirmed no other workflow, Makefile, or script referenced the removed files


mailo-nr (approved):
Nice!

PR #629, opened 2026-07-23
ci: remove source caching, fix goreleaser cache and ensure fips sources only generate for fips

agarvin-nr:
### Summary
* Splits source generation to only make `_build` or `_build-fips` depending on `${{ inputs.fips }}`, not both
* Removes source caching:
  * The above split reduced the source generation time to 30s per run
  * No time was being saved on cache, as source generation is required for license validation
  * This fixes a race condition that was impacting GoReleaser caching (see **Evidence**), which will ultimately save anywhere from 16 to 30 minutes per PR
* Removes `build.log` from `_build-*` directories prior to GoReleaser caching
  * This file contains timestamps which would change the cache key every run, and is unused for artifact generation
* Reorganizes the `scripts` directory into `scripts/build`, `scripts/misc` and `scripts/release`

### Validation
**Updated**
In this PR, all against commit `de56f52`
* [Run 1](https://github.com/newrelic/nrdot-collector-releases/actions/runs/30404038601/job/90425293762?pr=629): Deleted cache prior to run, all fresh builds. Source generation + all checks take 40s
* [Run 2](https://github.com/newrelic/nrdot-collector-releases/actions/runs/30404038601/job/90429593487?pr=629): Cache hit for GoReleaser build

### Evidence
There is a race condition on source caching where `nrdot-collector` and `nrdot-collector-fips` write to same sources cache key despite producing different source directories. This has the following negative downwind effects on GoReleaser builds:

**Run 1:** ([bf5382c](https://github.com/newrelic/nrdot-collector-releases/actions/runs/30044318308))
* **Source Generation**: Files are generated from scratch. `nrdot-collector` and `nrdot-collector-fips` separately invoke `make ci`, which builds both `_build` and `_build-fips` directories irrespective of fips status. Each OCB run creates `build.log` files with different timestamps. Race condition: Both distros write to same source cache key, `nrdot-collector-fips` is faster and writes its version to cache.
* **GoReleaser Generation**: Binaries are generated from source files created in this run. Cache keys are generated as such:
  * `nrdot-collector-{hashed _build with timestamp X}`
  * `nrdot-collector-fips-{hashed _build with timestamp Y }`

**Run 2:** ([49cce16](https://github.com/newrelic/nrdot-collector-releases/actions/runs/30046207132))
* **Source Generation**: Cache hit from Run 1, both `nrdot-collector` and `nrdot-collector-fips` retrieves source files with timestamp Y
* **GoReleaser Generation**: `nrdot-collector` seeks cache like `nrdot-collector-{hashed _build with timestamp Y}` which cannot exist due to race. Cache misses and `nrdot-collector` unnecessarily rebuilds goreleaser. Saves to new cache key `nrdot-collector-{hashed _build with timestamp Y}`

**Run 3:** ([39fae7a1](https://github.com/newrelic/nrdot-collector-releases/actions/runs/30054231713))
* Same source file cache hit as Run 2
* Now `nrdot-collector` has a cache hit for `nrdot-collector-{hashed _build with timestamp Y}`

The fix for this is removing `build.log` from the cache, AND either fixing the source cache key to use `${FIPS}` or to remove caching. The latter option makes things much easier, as caching does not save very much time.


scripts/build/build.sh
```diff
@@ -2,7 +2,7 @@
 # Copyright New Relic, Inc. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-REPO_DIR="$( cd "$(dirname "$( dirname "${BASH_SOURCE[0]}" )")" &> /dev/null && pwd )"
+REPO_DIR="$(git rev-parse --show-toplevel)"
```
agarvin-nr, line 5: The original doesn't work with the new directory structure. Instead of using `dirname(dirname(dirname()))`, I took this `git rev-parse` pattern from our [makefile](https://github.com/newrelic/nrdot-collector-releases/blob/main/Makefile).

.github/workflows/ci-base.yaml
```diff
@@ -96,11 +96,12 @@ jobs:
...
           SOURCES_HASH="${{ hashFiles(
             format('distributions/{0}/manifest.yaml', inputs.distribution),
-            'scripts/build.sh',
+            'scripts/build/**',
             'Makefile',
             'fips/fips.go'
           ) }}"
-          echo "sources_key=sources-${{ inputs.distribution }}-${SOURCES_HASH}" >> $GITHUB_OUTPUT
+          FIPS="${{ inputs.fips == true && '-fips' || '' }}"
+          echo "sources_key=sources-${{ inputs.distribution }}${FIPS}-${SOURCES_HASH}" >> $GITHUB_OUTPUT
```
kb-newrelic: Great find, that's a pretty subtly bug

Makefile
```diff
@@ -43,10 +43,10 @@ ensure-goreleaser-up-to-date: generate-goreleaser
 	@git diff -s --exit-code distributions/*/.goreleaser*.yaml || (echo "Check failed: The goreleaser templates have changed but the .goreleaser.yamls haven't. Run 'make generate-goreleaser' and update your PR." && exit 1)
 
 validate-components:
-	@./scripts/validate-components.sh
+	@./scripts/build/validate-component-inventory.sh
```
kb-newrelic: I guess this actually shouldn't invalidate the build cache and instead just run as a separate step in CI?
Looks like there is already the `check` target that accomplishes that goal. How about we move this target from `ci` to `check` and also add the `validate-actions-hashes` as a dependency to check remove that as a separate step from CI?
`validate-component-inventory.sh` should then also no longer be under `./scripts/build` and instead live under `misc`

agarvin-nr: That sounds better! I added all of those checks as dependencies to `make check`.

agarvin-nr: I also moved `licenses-check` down into check, since it should probably run every time regardless of source cache hit

Makefile
```diff
@@ -18,35 +18,33 @@ GO_LICENCE_DETECTOR_CONFIG   := $(SRC_ROOT)/internal/assets/license/rules.json
...
-check: ensure-goreleaser-up-to-date
+ci: pre-check generate-sources post-check
 
-build: go ocb
-	@./scripts/build.sh -d "${DISTRIBUTIONS}" -b ${OTELCOL_BUILDER}
+pre-check: goreleaser-file-check manifests-check component-inventory-check actions-hashes-check
 
-build-fips: go ocb
-	@./scripts/build.sh -d "${DISTRIBUTIONS}" -b ${OTELCOL_BUILDER} -f true
+generate-sources: go ocb
```
kb-newrelic: nit: the original split between `generate-sources` and `build` was because `build` would both generate sources and compile whereas the former did just the first. I wonder if using the word `build` is more intuitive than `generate-sources` now that we no longer make that distinction.

kb-newrelic (dismissed):

Makefile
agarvin-nr: I agree! I made this change

kb-newrelic (dismissed):

kb-newrelic (approved):

PR #636, opened 2026-07-30
chore: add claude skill to clean up stale replace directives

agarvin-nr:
### Summary
* Adds a claude skill to clean up any replace directives which are stale (e.g. modules whose current versions exceed the fixed versions)
* Cleans up stale CVE replaces using the new skill
* Fixes an issue was causing the `version-bump-analysis.sh` script to exit with a SIGPIPE error, which would delay / confuse Claude.
  * I believe it was caused by `awk` exiting the pipe before `curl` had a chance to complete / pipe its fetch. I'm not totally sure why this started happening - It's possible that the upstream changelog is large enough now that it can't cleanly send its contents before `awk` exits. Removing the pipe seems to have fixed it.
 
### Validation
* Claude skill used to remove these replace directives
* New version bump script was used to generate [v0.157.0 analysis](https://github.com/newrelic/nrdot-collector-releases/pull/643#issuecomment-5146918350)


.claude/skills/fix-cve/skill.md
```diff
@@ -122,24 +122,14 @@ replaces:
   - github.com/prometheus/prometheus v0.311.2-0.20260409145810-72293ff1d2e0 => github.com/prometheus/prometheus v0.311.3
 ```
 
-### 6. Regenerate licenses
```
agarvin-nr, line 125: CLAUDE.md has a directive to make licenses. I noticed it claude automatically gens licenses in the new skill, so I believe this step is no longer necessary.

.claude/skills/cleanup-cve/skill.md
```diff
@@ -0,0 +1,49 @@
...
+
+The current version of any given module comes from from the go.sum files generated by OCB. Build source files for each distribution first (even if distribution/*/_build exists, as it may be out of date):
+
+```bash
+make build
+```
+
+### 2. Seek replace directives and find current versions
+
+We need to determine which replace directives are stale in the current version of NRDOT. Find all replaced modules and, determine if the current go.sum version exceeds or equals the fixed version (e.g. right side of =>).
```
kb-newrelic: If a replace actually took place, would the version not be equal, so the directive should stay? Or rather you wouldn't be able to tell whether the replace took place or not?

agarvin-nr: Oh shoot, you're right! I had it exceeds-only initially, I'm not sure how I convinced myself about changing it to >=. I'll change this back to exceeds-only, good catch!

agarvin-nr: I fixed it and re-ran the skill, this PR removes two of the three replace directives now.

PR #638, opened 2026-07-30
chore: test

agarvin-nr:
DO NOT MERGE

### DIFF
--- a/scripts/build/validate-source-files.sh
+++ b/scripts/build/validate-source-files.sh
@@ -55,3 +55,5 @@ else
     echo "❌ files not found: ${missing_files[*]}"
     exit 1
 fi
+
+echo "nothing"


agarvin-nr:
Result of this test: First run of every PR will still always result in a cache miss for built GoReleaser files. This is for two reasons:

* Primary: `inputs.publish` is false on main. This is (partially?) desired, since we need main CI to run on a fresh build every time, but it also means there there it won't save a cache.
* Secondary: Even if main saved a cache, the goreleaser arguments are cached based on its arguments. Main does not use the `--snapshot` flag, so its key would differ every time.

This means some kinds of PRs (like script or tooling changes) on files which aren't already [excluded](https://github.com/newrelic/nrdot-collector-releases/blob/main/.github/workflows/ci.yaml#L9-L17) cannot benefit from main's cache. As each GoReleaser run takes ~16m, we might want to reconsider this behavior.

